PR #1478, opened 2025-02-03
Subxt tests

haider-rs:
- Adds local cache to store data for chronicle transactions.
- Add Mock implementation for timechain communication.
- Writes test for the communication between subxt and timechain.


dvc94ch (commented):
Nice. Although I think all the calls that work offline, like signing, you don't have to mock

dvc94ch (commented):
Like sign_payload for example...

dvc94ch (approved):

primitives/src/shard.rs
```diff
@@ -23,8 +23,83 @@ pub type TssHash = [u8; 32];
...
+	}
+
+	pub fn deserialize<'de, D>(deserializer: D) -> Result<ProofOfKnowledge, D::Error>
+	where
+		D: Deserializer<'de>,
+	{
+		let bytes = <Vec<u8>>::deserialize(deserializer)?;
+		ProofOfKnowledge::try_from(bytes).map_err(|_| D::Error::custom("invalid public key length"))
+	}
+}
```
dvc94ch: why are these needed?

tc-subxt/Cargo.toml
```diff
@@ -28,6 +28,11 @@ subxt-signer.workspace = true
 
 time-primitives = { workspace = true, default-features = true }
 tc-subxt-metadata = { path = "metadata" }
+env_logger.workspace = true
+tokio-stream = { version = "0.1", features = ["sync"]}
+redb = "2.1.2"
+bincode = "1.3.3"
+
```
dvc94ch, line 36: would be nice to keep them ordered alphabetically, also why is env_logger needed here?

dvc94ch: ah, I assume it should be a dev-dependency?

primitives/src/shard.rs
dvc94ch: I see now, you're using serde + bincode instead of parity scale codec

tc-subxt/src/db.rs
```diff
@@ -0,0 +1,95 @@
...
+		write_tx.commit()?;
+		Ok(Self { db, public_key })
+	}
+}
+
+impl ITransactionDbOps for TransactionsDB {
+	fn store_tx(&self, tx_data: &TxData) -> Result<()> {
+		let mut composite_key = [0u8; 64];
+		composite_key[..32].copy_from_slice(&self.public_key);
+		composite_key[32..].copy_from_slice(tx_data.hash.as_bytes());
```
dvc94ch, line 33: wouldn't it be easier to index txdata by the nonce instead?

tc-subxt/src/lib.rs
```diff
@@ -11,16 +12,20 @@ use subxt::backend::rpc::RpcClient;
...
-mod worker;
+pub mod db;
+pub mod metadata;
+pub mod timechain_client;
+pub mod worker;
 
 use metadata::technical_committee::events as CommitteeEvent;
 
 pub use subxt_signer::sr25519::Keypair;
+pub const DB_PATH: &str = "cached_tx.redb";
```
dvc94ch: path should be provided as an argument

dvc94ch (approved):
looks good

tc-subxt/src/timechain_client.rs
```diff
@@ -0,0 +1,239 @@
...
+
+#[async_trait::async_trait]
+pub trait ITimechainClient {
+	type Submitter: ITransactionSubmitter + Send + Sync;
+	type Block: IBlock + Send + Sync;
+	type Update: Send + Sync;
+	async fn get_latest_block(&self) -> Result<BlockDetail>;
+	fn sign_payload<Call>(&self, call: &Call, params: ExtrinsicParams) -> Vec<u8>
+	where
+		Call: Payload + Send + Sync;
```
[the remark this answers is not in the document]
haider-rs: its easier to debug problems with current imlementation since its test the hash can be easily mapped to the nonce of user.

tc-subxt/src/db.rs
haider-rs: actually if a chronicle would change the account of the chronicle then the transactions in the db will not be for that account. thats why i mapped with the public key instead.

PR #1480, opened 2025-02-04
[WIP] Support native token teleportation to ERC-20

Lohann:
## Description

This PR introduces basic support to reserve-based-transfers.

## Timechain to external ERC20:
1 - The account calls `assets_bridge.teleport_keep_alive(...)`, and specify the network and amount of tokens to be teleported..
2 - The pallet lock the user's funds in the bridge pot (account managed by the pallet).
3 - A new GMP task is pushed to the queue.

## ERC20 to Timechain:
1 - The account calls the `ERC20.teleport(..)` in the source chain.
2 - The ERC20 contract calls the Gateway Contract, which emits a new message to the timechain RouteI/Network id.
3 - A Chronicle feed that event to the timechain.
4 - The timechain have a special handler for GMP messages which uses timechain network id, this message if valid is handled by pallet-assets-bridge.
5 - The pallet-assets-bridge unlock the tokens and transfer to the specified account. 

## TODOs
+ [X] Timechain to external ERC20
   done in #1491 
+ [x] ERC20 to Timechain
  #1499 
+ [X] make tc-cli do `setRoute` for TC nw
+ [X] `AnlogTokenV2`
  done in https://github.com/Analog-Labs/erc20-token/pull/17
+ [x] Automated e2e tests 
   - [x] [tc->erc20](https://github.com/Analog-Labs/timechain/pull/1580)
   - [x] erc20->tc
+ [ ] Refactor


dvc94ch (commented):
I have no idea what this is. There is a whole bunch of stuff that is just copied from other crates like the benchmarking, mock, tests. There are no tests. The pallet isn't registered in the runtime. If avneet signs of on it we can merge cc @foravneet , since he commissioned this work he should understand what was done.

dvc94ch:
@agryaznov is this still relevant?

agryaznov:
> @agryaznov is this still relevant?

yes, it is wip

node/src/chain_spec.rs
```diff
@@ -85,7 +89,7 @@ impl Default for GenesisKeysConfig {
 				Alice.to_raw_public().unchecked_into(),
 			)],
 			chronicles: vec![],
-			endowments: vec![],
+			endowments: vec![(BRIDGE_PALLET_ID.into_account_truncating(), 1)],
```
dvc94ch, line 92: why is there one token in here? don't think that's even enough for the existential deposit

node/src/chain_spec.rs
```diff
@@ -256,6 +261,9 @@ impl GenesisKeysConfig {
 			"technicalCommittee": {
 				"members": Some(self.admins.clone()),
 			},
+			"sudo": {
+				"key": root_account,
+			}
```
dvc94ch, line 266: we removed sudo, doubt this will fly cc @FlorianFranzen

dvc94ch: maybe check how the technical comittee works in various pallets, its quite easy to use that instead

config/envs/local/local-e2e-bridge.yaml
```diff
@@ -0,0 +1,30 @@
...
+    batch_offset: 0
+    batch_gas_limit: 10000000
+    gmp_margin: 0.0
+    shard_task_limit: 50
+    route_gas_limit: 10000000
+    route_base_fee: 1400000000
+    shard_size: 1
+    shard_threshold: 1
+chronicles:
+  - http://localhost:8080
```
dvc94ch, line 30: why do we need yet another config?

docker-compose.yml
```diff
@@ -99,6 +103,10 @@ services:
     environment:
       RUST_LOG: 'tc_subxt=debug,chronicle=debug,tss=debug,gmp_evm=info'
       RUST_BACKTRACE: 1
+    volumes:
+        - ${PWD}/tmp/chron/:/etc/keys/:ro
```
dvc94ch, line 107: why isn't anvil/state a volume? the keyfiles should probably be distributed with the anvil state to make sure they match

docker-compose.yml
```diff
@@ -99,6 +103,10 @@ services:
...
+    ports:
+    - 8080:8080
```
dvc94ch, line 109: port forwarding causes issues as discussed previously

pallets/tasks/src/lib.rs
```diff
@@ -49,9 +49,24 @@ pub mod queue;
...
+/// Teleport handlers.
+pub trait TeleportReciever<T> {
+	fn handle_teleport(recipient: &AccountId, amount: Balance) -> DispatchResult;
+}
+
+impl<T> TeleportReciever<T> for () {
+	fn handle_teleport(_recipient: &AccountId, _amount: Balance) -> DispatchResult {
+		Ok(())
+	}
+}
```
dvc94ch, line 64: what's the purpose of this trait?

pallets/tasks/src/lib.rs
```diff
@@ -474,6 +491,22 @@ pub mod pallet {
...
+					{
+						use polkadot_sdk::sp_core::U256;
+						let msg_id = msg.message_id();
+						if msg.bytes.len() == 96 {
+							let mut recipient = time_primitives::Address::default();
+							recipient.copy_from_slice(&msg.bytes[32..64]);
+							let amount = U256::from_big_endian(&msg.bytes[64..]).low_u128();
+							let _ = T::Teleporter::handle_teleport(&recipient.into(), amount);
+						} else {
+							Self::ops_queue(msg.dest_network).push(GatewayOp::SendMessage(msg));
```
dvc94ch, line 505: why are invalid messages queued? should never queue a message if dest_network == tc_network()

pallets/tasks/src/lib.rs
```diff
@@ -474,6 +491,22 @@ pub mod pallet {
...
+						}
```
dvc94ch, line 506:
```suggestion
if msg.est_network == T::Netoworks::tc_network() {
						if msg.bytes.len() == 96 {
							let mut recipient = time_primitives::Address::default();
							recipient.copy_from_slice(&msg.bytes[32..64]);
							let amount = U256::from_big_endian(&msg.bytes[64..]).low_u128();
							let _ = T::Teleporter::handle_teleport(&recipient.into(), amount);
						} else {
						log::error!("received invalid bridge message");
						}
```

pallets/tasks/src/lib.rs
```diff
@@ -474,6 +491,22 @@ pub mod pallet {
...
 						ShardRegistered::<T>::remove(pubkey);
 					},
+					GmpEvent::MessageReceived(msg)
+						if msg.dest_network.eq(&T::Networks::tc_network()) =>
+					{
+						use polkadot_sdk::sp_core::U256;
+						let msg_id = msg.message_id();
+						if msg.bytes.len() == 96 {
+							let mut recipient = time_primitives::Address::default();
+							recipient.copy_from_slice(&msg.bytes[32..64]);
```
dvc94ch, line 501: sender isn't checked, the sender must be the bridge address

runtime/src/configs/custom.rs
```diff
@@ -96,13 +103,90 @@ impl pallet_networks::Config for Runtime {
...
+			bytes: teleport_command.to_vec(),
+		};
+
+		data.incr_nonce()
+			.ok_or(pallet_assets_bridge::Error::<Runtime>::NetworkNonceOverflow)?;
+		Self::push_gmp_message(msg);
+
+		Ok(())
+	}
+}
```
dvc94ch, line 182: why aren't these methods part of the asset bridge pallet?

tc-cli/src/lib.rs
```diff
@@ -681,6 +688,24 @@ impl Tc {
...
+			gateway: src_gateway,
+			// TODO ??explain
+			relative_gas_price: (1, 1),
+			// TODO ??explain
+			gas_limit: 15_000_000,
+			base_fee: Default::default(),
+		};
+		self.println(None, format!("set_tc_route {:?}", &route)).await?;
+		connector.set_route(src_gateway, route).await
+	}
```
dvc94ch, line 708: this is unnecessary, already implemented.

tc-cli/src/lib.rs
```diff
@@ -46,9 +46,14 @@ pub struct Tc {
 	runtime: SubxtClient,
 	connectors: HashMap<NetworkId, Arc<dyn IConnectorAdmin>>,
 	msg: Sender,
+	tc_network_id: NetworkId,
 }
 
 impl Tc {
+	pub fn runtime(&self) -> &SubxtClient {
+		&self.runtime
+	}
```
dvc94ch, line 55: not sure why this needs to be public

tc-subxt/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub type ExtrinsicParams =
 	<DefaultExtrinsicParams<PolkadotConfig> as subxt::config::ExtrinsicParams<PolkadotConfig>>::Params;
 
 pub struct SubxtClient {
-	client: OnlineClient,
+	pub client: OnlineClient,
```
dvc94ch, line 38: also doesn't need to be pub

dvc94ch (commented):
haven't reviewed the asset-bridge or the tests yet

pallets/assets-bridge/README.md
```diff
@@ -0,0 +1,229 @@
...
+   2. Query [GmpStatus](https://github.com/Analog-Labs/analog-gmp-examples/blob/00090ef5b83574c5fdaa2a10d428f87e1702cc79/examples/teleport-tokens/BasicERC20.sol) of the message  via querying `gmpInfo(bytes32)` on the gateway:
+      
+      !note: even if tc-cli report task status as __completed_, message itself could have been reverted. 
+      To check actual message status run: 
+      ```sh
+      cast call 0x49877F1e26d523e716d941a424af46B86EcaF09E "gmpInfo(bytes32)" <msg_id>
+      ``` 
+   2. Query GMP message tx_hash by batch_id from tasks pallet storage: `batchTxHash(u64)`;
+   3. Re-run that tx with trace: 
+      `$> cast run <tx_hash>`;
```
dvc94ch, line 228: most of these things should be automated in tc-cli...

node/src/chain_spec.rs
agryaznov: this actually puts exactly 1 ANLOG (which is ED)

node/src/chain_spec.rs
agryaznov: yeah will do once we've got basic flow working and covered with e2e tests. sudo was done for the ease of demo as well

config/envs/local/local-e2e-bridge.yaml
agryaznov: 1. tests run outside containers, we cant' re-use local-evm config which specifies ulrs using docker container hostnames
2. we need only 1 external chain, whereas other configs specify 2 of them

docker-compose.yml
agryaznov: because connector derives key in a different way than anvil does...

docker-compose.yml
agryaznov: any suggestions on how to fix that?

runtime/src/configs/custom.rs
agryaznov: Loosely coupled design is better. any pallet implementing `AssetTeleporter` could be the teleport handler

node/src/chain_spec.rs
dvc94ch: analog tokens have 12 decimal places. 1 ANLG would be written as `1 * ANLG` where `const ANLG = 1e12`

tc-cli/src/lib.rs
agryaznov: great if so, could be updated to use the existing impl then. Could you pls specify where is it?

docker-compose.yml
dvc94ch: maybe run cargo test in docker? also there are custom test runners you can configure for running tests on qemu or embedded systems. maybe we need to look into that? however I'd prefer we make that a separate PR...

tc-cli/src/lib.rs
agryaznov: in [tests we need ](https://github.com/Analog-Labs/timechain/blob/4e9a2a5ca8b9d1ee7c1e2a5f7b370013dedfb583/e2e-tests/tests/bridge.rs#L72)to get client (`OnlineClient`)

tc-subxt/src/lib.rs
agryaznov: same as above

tc-cli/src/lib.rs
dvc94ch: that should be wrapped in `tc-subxt`.
1. having it in one place prevents code duplication
2. makes tests conciser and easier to read
3. contains build errors due to metadata changes in a single place.
4. if tc-cli parallelizes txs we'll need the tx manager in tc-subxt (we did quite a bit of work on that recently)

tc-cli/src/lib.rs
dvc94ch: test_setup does that automatically. just provide the networks you need to setup. its handled in a deploy step where the routes are updated to match prices.csv and config.yaml

dvc94ch: so in practice, make sure test-setup or deploy is called and prices.csv and config.yaml are configured appropriately

runtime/src/configs/custom.rs
dvc94ch: well, loose coupling aften adds to system complexity and hinders readability. I don't like adding new traits unless there are at least two implementations of it.

I understand that substrate enforces this trait mess on us. so the way we handle this is by adding a trait for a pallet [here](https://github.com/Analog-Labs/timechain/blob/4e9a2a5ca8b9d1ee7c1e2a5f7b370013dedfb583/primitives/src/lib.rs#L210) which specifies the pallet interface (public methods for other pallets) and the pallet is added to the config keeping the same name and bounded by AssetBridgeInterface

docker-compose.yml
dvc94ch: don't understand exactly what this has to do with it. but seems like this is some bug we need to fix. I assume that the connectors derive the keys how metamask would derive them for a network, compliant with bip44/bip32/bip39. if that is not the case we need to fix it. then on anvil, the prefunded key is used by the faucet in the connector, and then it doesn't matter anymore how anvil derives anything anyway

agryaznov:
integration tests are [done](https://github.com/Analog-Labs/timechain/pull/1586); the work on this could be resumed later on

PR #1516, opened 2025-02-07
primitives: properly integrate with ss58 registry

FlorianFranzen:
## Description

This ensures that the correct ss58 prefix is used at all times by properly configuring sp_core where needed.


penumbra23:
@FlorianFranzen So this initializes the clients to always use the proper ss58 format based on the metadata I suppose? No need to manually set it anymore?

FlorianFranzen:
It sets some of the internal defaults and the IDs come from the official (and patched) registry.

dvc94ch (approved):
not sure why this is better or why ci is acting up...

peektism (approved):

FlorianFranzen:
Will merge it as soon as I can get a clean integration test run.

PR #1538, opened 2025-02-10
Migration to remove buggy pool members

### LINKED ISSUE #1502: Fix NominationPools failed try_state check
Current CI try-runtime error is due to a failing migration for the NominationPools pallet:
```
[2025-02-06T17:46:49Z ERROR runtime::frame-support] ❌ "NominationPools" try_state checks failed: Other("No member should have zero points")
```

peektism:
Closes https://github.com/Analog-Labs/timechain/issues/1502


peektism:
@FlorianFranzen if I write this migration correctly will the try-runtime check pass?

FlorianFranzen:
That would be the goal. It still fails with the following:
```
[2025-02-11T20:26:13Z ERROR try-runtime] Detected errors while executing `try_state`:
[2025-02-11T20:26:13Z ERROR try-runtime] Other("No member should have zero points")
[2025-02-11T20:26:13Z ERROR runtime] panicked at /home/runner-3/runner/_work/timechain/timechain/runtime/src/apis.rs:546:65:
    called `Result::unwrap()` on an `Err` value: Other("Detected errors while executing `try_state` checks. See logs for more info.")
thread 'main' panicked at cli/main.rs:326:10:
called `Result::unwrap()` on an `Err` value: Input("failed to execute TryRuntime_on_runtime_upgrade: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n    0: 0x4221b - timechain_runtime.wasm!rust_begin_unwind\n    1: 0x91cd - timechain_runtime.wasm!core::panicking::panic_fmt::h92fad10c4ee47658\n    2: 0x842a - timechain_runtime.wasm!core::result::unwrap_failed::h02a37653da911315\n    3: 0x2668c3 - timechain_runtime.wasm!TryRuntime_on_runtime_upgrade")
```
Seems whatever you are doing is not working yet, as you are still hitting [this check](https://github.com/Analog-Labs/polkadot-sdk/blob/1b04a0ba68dc4cd979cdd9f2e3ce6aed67d5db18/substrate/frame/nomination-pools/src/lib.rs#L3646C15-L3646C27).

runtime/src/migrations/fix_pool_members.rs
```diff
@@ -0,0 +1,103 @@
...
+				0x32, 0x92, 0xd8, 0x61, 0xff, 0xc3, 0xb9, 0x23, 0x45, 0x7d, 0x68, 0x91, 0xa2, 0x5e,
+				0x4a, 0xb3, 0x67, 0xc2,
+			])
+			.into(),
+		];
+
+		for account in accounts_to_remove.iter() {
+			if PoolMembers::<T>::contains_key(account) {
+				PoolMembers::<T>::remove(account);
+				weight += <T as frame_system::Config>::DbWeight::get().writes(1);
```
FlorianFranzen: A `log::info!("Removed pool member: {account:?}");` here should let us know if try-runtime actually does anything.

libreamartists: It did not seem to do anything, maybe it is not applying the migration to storage?

FlorianFranzen:
These accounts should be more something like this:
```
hex2array!("183982ce80e4b52f2e80aaf36d18b1eba1a32005ffbefd952962227f2f4db309") 
hex2array!("aeedb1738fb7133da240213ab0f0916eb7abaf4140693b7a3f54311732bdda33")
hex2array!("b20754dceabb4f4e9fa4b8923711a87ddb263368750db3a8f05e4e5e5ea5a155")
hex2array!("aa8703194b1b555a1dc1ba17b099467db45797444e9cc4aecf2a5f95d275b57a")
```
Or just parse them in the runtime, for four accounts it is probably fine.

FlorianFranzen:
Ok, progress: `Each `BondedPool.member_counter` must be equal to the actual count of members of this pool`.

peektism:
Ok updating BondedPool.member_count in the migration accordingly, thanks

peektism:
More progress: `[2025-02-12T17:18:05Z ERROR runtime::frame-support] ❌ "NominationPools" try_state checks failed: Other("TVL must be equal to or less than the total balance of all PoolMembers.")`

Working on updating TVL now

peektism:
Last commit did not fix the error. Still debugging `[2025-02-13T16:07:43Z ERROR runtime::frame-support] ❌ "NominationPools" try_state checks failed: Other("TVL must be equal to or less than the total balance of all PoolMembers.")`

runtime/src/migrations/fix_pool_members.rs
```diff
@@ -0,0 +1,93 @@
...
+		// Track affected pools to update member_counter
+		let mut affected_pools =
+			sp_std::collections::btree_map::BTreeMap::<u32, (u32, BalanceOf<T>)>::new();
+
+		for account in &accounts_to_remove {
+			if let Some(member) = PoolMembers::<T>::take(account) {
+				log::info!("Removed pool member: {account:?}");
+				weight += T::DbWeight::get().writes(1);
+				let pool_id = member.pool_id; // Extract the pool ID correctly
+				let stake_amount = member.points; // Get the stake amount in raw balance
```
libreamartists: may need to convert points to balances here, I had previously assumed it was 1:1 which is what it is at first but if any slashing has occurred then this assumption is incorrect

runtime/src/migrations/fix_pool_members.rs
```diff
@@ -0,0 +1,93 @@
...
+						pool.member_counter = 0;
+					}
+
+					// Ensure TVL does not go negative
+					// if pool.points >= *removed_stake {
+					// 	pool.points -= *removed_stake;
+					// } else {
+					// 	pool.points = Zero::zero();
+					// }
+					TotalValueLocked::<T>::mutate(|x| *x -= *removed_stake);
```
libreamartists: without this line, we get the error `TVL must be equal to or less than the total balance of all PoolMembers.` but with it we get the error `TVL deviates from the actual sum of funds of all Pools.` so there may be other storage items or parts of the storage item that need to be updated

libreamartists: I removed the points check above to see how that would effect things but I doubt it will resolve this problem

peektism:
Still getting this error `TVL deviates from the actual sum of funds of all Pools.` so I think we need to keep that check but also decrease the balance of the Pool itself. I am looking into it but would really appreciate a second pair of eyes on this so that I don't keep trying things that do not work cc @FlorianFranzen @dvc94ch @agryaznov

peektism:
Seems like we need to update `T::StakeAdapter::total_stake(Pool::from(bonded_pool.bonded_account()))` so that the sum for all bonded_pools equals the TotalValueLocked after it is decreased. I am still not sure how that may updated, it might require adjusting the balance for the `bonded_pool.bonded_account()`.

FlorianFranzen:
I thought the whole issue is that their balance is zero, so how can they affect the sum?

peektism:
> I thought the whole issue is that their balance is zero, so how can they affect the sum?

I do not know.

Would really appreciate a second pair of eyes on this before trying anything else.

```
[2025-02-13T18:28:13Z WARN  runtime::nomination-pools] [736046] 🏊‍♂️ reward pool of 1: 0 (ed = 1000000000000), should only happen because ED has changed recently. Pool operators should be notified to top up the reward account
[2025-02-13T18:28:13Z ERROR runtime::frame-support] ❌ "NominationPools" try_state checks failed: Other("TVL deviates from the actual sum of funds of all Pools.")
```

peektism:
As discussed during SP, @FlorianFranzen will commit to this branch a fix which will only keep the first two accounts that are being removed from the pool. This should resolve the try-runtime error.

peektism:
We may want to open a follow up issue related to the other two buggy accounts which do have a balance but apparently they have other problems, not so sure of specifics

FlorianFranzen:
@4meta5 Alright, added the changed code, wrapped it in a storage migration to avoid any issues during deployment. Please give it a quick review, then I will deploy it asap.

runtime/src/migrations/fix_pool_members.rs
```diff
@@ -0,0 +1,90 @@
...
+				let account: T::AccountId = AccountId32::from_ss58check(account)
+					.map_err(|_| "Failed to parse address")?
+					.into();
+
+				if let Some(member) = PoolMembers::<T>::take(&account) {
+					// Track weight of pool members update
+					weight += T::DbWeight::get().reads_writes(1, 1);
+
+					// Ensure this is actually an offending account
+					ensure!(member.points == Zero::zero(), "Member points not zero");
```
libreamartists, line 54: is it fine to panic in a migration? won't that brick the chain if we ask all clients to upgrade and then the migration panics?

peektism (commented):
Approved lgtm

runtime/src/migrations/fix_pool_members.rs
FlorianFranzen: `ensure` just returns an error here that is caught by the storage layer. That way any single error here leads to a full revert of everything done in the storage layer.

We actually use a `Result<(), &str> here to easily pass error message through the storage layer.

PR #1539, opened 2025-02-11
eth-bridge: enable SORA eth-bridge for mainnet

### LINKED ISSUE #1515: Runtime-specific weights for SORA pallets
ideally weights computed in context of runtime but this is fine for right now

_Originally posted by @4meta5 in https://github.com/Analog-Labs/timechain/pull/1514#discussion_r1947073469_

dvc94ch:
Closes #1515


runtime/src/configs/bridge.rs
```diff
@@ -46,5 +46,5 @@ impl bridge_multisig::Config for Runtime {
 	type DepositBase = DepositBase;
 	type DepositFactor = DepositFactor;
 	type MaxSignatories = MaxSignatories;
-	type WeightInfo = bridge_multisig::weights::SubstrateWeight<Runtime>;
+	type WeightInfo = weights::bridge_multisig::WeightInfo<Runtime>;
```
libreamartists, line 49: should this be set based on testnet vs mainnet weights used?

penumbra23: It does so by selectively importing here in the `runtime/src/weights/mod.rs` based on the feature flag given (`testnet` or default)

penumbra23:
As usual, `try-runtime` fails the check (because of the NominationPools issue we have) but it compiles

peektism (approved):

FlorianFranzen (approved):
Looks like the code activated successfully: 
```
🐥 New pallet "EthBridge" detected in the runtime. Initializing the on-chain storage version to match the storage version defined in the pallet: StorageVersion(1)
🐥 New pallet "BridgeMultisig" detected in the runtime. The pallet has no defined storage version, so the on-chain version is being initialized to StorageVersion(0).
```
Anything these two pallets do to our runtime logic we should be aware about?

FlorianFranzen:
@dvc94ch Hope it was ok that I added the node side of things as well.

FlorianFranzen:
Ok, looks like it now also builds a bridge node for us. Gave it a quick try, seems ready to go.

PR #1540, opened 2025-02-11
Refactor estimate gas cost.

### LINKED ISSUE #1416: log/measure tx/msg cost
- all tx costs should be logged in tc-subxt/backend
- smoke test and benchmark should display message cost and average message cost
- benchmark should display average message latency

dvc94ch:
Closes #1416


haider-rs (approved):
LGTM!

PR #1544, opened 2025-02-11
[env] dev to use anvil instead of geth for evm chains

### LINKED ISSUE #1543: tc-cli should support debugging target chain transactions

### LINKED ISSUE #1548: onGmpReceived should provide the nonce as an argument
is useful for ordering messages and reconstructing the GmpMessage struct in the GmpProxy.

agryaznov:
Closes #1543 
Closes #1548


docker-compose.yml
```diff
@@ -76,26 +76,12 @@ services:
...
-    - '--http.vhosts=*'
-    - '--http.corsdomain=*'
-    - '--http.api=eth,debug,admin,txpool,web3,net'
-    - '--ws'
-    - '--ws.addr=0.0.0.0'
-    - '--ws.port=8545'
-    - '--ws.origins=*'
-    - '--ws.api=eth,debug,admin,txpool,web3,net'
-    - '--verbosity=4'
+    - 'anvil -b 2'
```
agryaznov: 2 secs, as `--dev.period=2` was set for geth

dvc94ch (approved):

dvc94ch:
Thx!

agryaznov:
Note this also would require some logic change reg. admin account funding, as connector's `faucet` woudn't work with anvil

```
2025-02-11T20:08:25.964414Z ERROR rosetta_server::ws::reconnect: rpc request 'eth_sendTransaction' failed: Call(ErrorObject { code: ServerError(-32003), message: "Insufficient funds for gas * price + value", data: None })
Call(
    ErrorObject {
        code: ServerError(
            -32003,
        ),
        message: "Insufficient funds for gas * price + value",
        data: None,
    },
)
```

dvc94ch:
So this needs some more work:
1. chain connector should be able to identify when dealing with anvil using the `admin_nodeInfo` rpc
2. chain connector should use `anvil_setBalance` in the faucet

dvc94ch:
never mind, the whole connector codebase is one giant hack. one more hack won't hurt...

dvc94ch (approved):

gmp/evm/Cargo.toml
```diff
@@ -16,11 +16,7 @@ async-trait.workspace = true
 futures.workspace = true
 hex.workspace = true
 reqwest = { version = "0.12.7", default-features = false, features = ["json", "rustls-tls-native-roots"] }
-rosetta-client = { git = "https://github.com/analog-labs/chain-connectors", rev = "37e2cd1" }
-rosetta-server = { git = "https://github.com/analog-labs/chain-connectors", rev = "37e2cd1" }
-rosetta-server-ethereum = { git = "https://github.com/analog-labs/chain-connectors", rev = "37e2cd1" }
-rosetta-ethereum-backend = { git = "https://github.com/analog-labs/chain-connectors", rev = "37e2cd1" }
-rosetta-ethereum-types = { git = "https://github.com/analog-labs/chain-connectors", rev = "37e2cd1" }
+rosetta-client = { git = "https://github.com/analog-labs/chain-connectors", rev = "84df6f5" }
```
agryaznov: I think this should be done a bit different, see my [comment](https://github.com/Analog-Labs/chain-connectors/pull/262#discussion_r1952610021)

agryaznov:
> So this needs some more work:
> 
>     1. chain connector should be able to identify when dealing with anvil using the `admin_nodeInfo` rpc
> 
>     2. chain connector should use `anvil_setBalance` in the faucet

I think this approach is better than a dirty hack proposed [here](https://github.com/Analog-Labs/chain-connectors/pull/262/commits/edc50319aff92087e4666940738c46cd8669c1b1) , with a minor changes: 
1. let's do it in chain connector by passing `blockchain: "anvil"` to params 
2. Then for anvil we pass private key of default dev account [here](https://github.com/Analog-Labs/timechain/blob/a076aaa7065d8497f2a1b463806d263edabd0a11/gmp/evm/src/lib.rs#L402)

gmp/evm/Cargo.toml
dvc94ch: the connector is already an unholy mess. have you taken a proper look at it yet? I plan to devote some serious time to changing that, but before launching gmp we won't have the time...

agryaznov: > I plan to devote some serious time to changing that, but before launching gmp we won't have the time...

exactly. that's the reason I suggest not to change it at all ("if it works somehow, don't touch it"), or if we really have to, then make those changes as minimal as possible; thus making needed changes for switching to anvil mostly on the backend side.

docker-compose.yml
```diff
@@ -76,26 +76,11 @@ services:
...
-    - '--http.api=eth,debug,admin,txpool,web3,net'
-    - '--ws'
-    - '--ws.addr=0.0.0.0'
-    - '--ws.port=8545'
-    - '--ws.origins=*'
-    - '--ws.api=eth,debug,admin,txpool,web3,net'
-    - '--verbosity=4'
+    - 'anvil -b 2 --steps-tracing --base-fee 0'
+    environment:
+        ANVIL_IP_ADDR: '0.0.0.0'
```
haider-rs, line 83: which port does it export the rpc to host machine for querying?

gmp/evm/src/lib.rs
```diff
@@ -375,15 +393,17 @@ impl IConnectorBuilder for Connector {
...
 		Self: Sized,
 	{
+		let (blockchain, private_key) = if params.blockchain == "anvil" {
+			let private_key = hex_literal::hex![
+				"ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+			];
+			(Blockchain::Ethereum, Some(private_key))
+		} else {
+			(params.blockchain.parse()?, None)
+		};
```
haider-rs, line 403: why do we need this?

haider-rs (approved):

docker-compose.yml
dvc94ch: none. but you can query it with `docker compose run tc-cli debug-transaction NETWORK_ID TX_HASH`

gmp/evm/src/lib.rs
dvc94ch: for faucet to work with anvil. the hard coded address is the prefunded anvil private key which is used by the faucet command in connector.

analog-gmp/src/interfaces/IGmpReceiver.sol
```diff
@@ -17,7 +17,7 @@ interface IGmpReceiver {
      * @param payload The message payload with no specified format
      * @return 32 byte result which will be stored together with GMP message
      */
-    function onGmpReceived(bytes32 id, uint128 network, bytes32 source, bytes calldata payload)
+    function onGmpReceived(bytes32 id, uint128 network, bytes32 source, uint64 nonce, bytes calldata payload)
```
dvc94ch, line 20: @agryaznov here's the change causing the issue

analog-gmp/test/utils/GmpProxy.sol
```diff
@@ -36,12 +37,22 @@ contract GmpProxy is IGmpReceiver {
...
+        unchecked { msgGasLimit = uint64(gasLimit + 579); }
+        GmpMessage memory message = GmpMessage({
+            source: src,
+            srcNetwork: uint16(srcNetwork),
+            dest: address(this),
+            destNetwork: NETWORK_ID,
+            gasLimit: msgGasLimit,
+            nonce: nonce,
+            data: payload
+        });
```
dvc94ch, line 55: here's the reason for the change.
1. nonce is determined by gateway
2. this was worked around by guessing the nonce before submitting the message and then encoding the GmpMessage struct in the payload.

gmp/evm/src/lib.rs
```diff
@@ -690,15 +697,48 @@ impl IConnectorAdmin for Connector {
...
+			source: src.into(),
+			nonce: 0,
+			payload: payload.into(),
+		};
+		let gas_limit = self
+			.wallet
+			.eth_send_call_estimate_gas(a_addr(contract).into(), call.abi_encode(), 0)
+			.await?;
+		Ok(gas_limit)
+	}
```
dvc94ch, line 720: we also guessed the gas_limit, which was hard coded to 300000 gas. this caused messages to be 5x overpriced

gmp/evm/src/lib.rs
```diff
@@ -690,15 +697,48 @@ impl IConnectorAdmin for Connector {
...
+			nonce: 0,
+			data: payload.into(),
+		};
 		let call = sol::Gateway::estimateMessageCostCall {
-			networkid: dest,
-			messageSize: U256::from(bytes.len()),
+			networkid: dest_network,
+			// abi_encoded_size returns the size without the 4 byte selector
+			messageSize: U256::from(msg.abi_encoded_size() + 4),
 			gasLimit: U256::from(gas_limit),
```
dvc94ch, line 742: this message used to be called internally in submit message, but we want to estimate the final cost before submitting the message

dvc94ch (commented):
cc @agryaznov

tc-cli/src/main.rs
```diff
@@ -336,44 +358,69 @@ async fn real_main() -> Result<()> {
...
-			msg.gas_cost = tc.estimate_message_cost(&msg).await?;
-			let msg_id = tc.send_message(msg).await?;
+			let payload = vec![];
+			let gas_limit = tc
+				.estimate_message_gas_limit(dest, dest_addr, src, src_addr, payload.clone())
+				.await?;
+			let gas_cost = tc.estimate_message_cost(src, dest, gas_limit, payload.clone()).await?;
+			let msg_id = tc
+				.send_message(src, src_addr, dest, dest_addr, gas_limit, gas_cost, payload.clone())
+				.await?;
```
dvc94ch, line 423: see here for how its used

tc-cli/src/main.rs
```diff
@@ -336,44 +358,69 @@ async fn real_main() -> Result<()> {
...
+			src_network,
+			dest_network,
+			gas_limit,
+			payload,
+		} => {
+			let payload = hex::decode(payload)?;
+			let gas_cost =
+				tc.estimate_message_cost(src_network, dest_network, gas_limit, payload).await?;
+			tc.println(None, gas_cost.to_string()).await?;
+		},
```
dvc94ch, line 386: and the new utilities we have in tc-cli

PR #1549, opened 2025-02-13
Retry failed tasks via admin

### LINKED ISSUE #1394: tc-cli should be able to list failed tasks and force the timechain to retry them

haider-rs:
## Description

- Saves failed tasks in a new storage in task pallet.
- Provides runtime api to get all failed tasks.
- Provides a new extrinsic to retry a failed task via admin.
- Provided tc-cli utils to list failed task.
- Provides tc-cli command to retry failed task.
- Adds tests in pallet task.
- Adds benchmarks for extrinsic `restart_batch`

Fixes # (issue)
closes: https://github.com/Analog-Labs/timechain/issues/1394

## Type of change

Please delete options that are not relevant.
- [x] New feature (non-breaking change which adds functionality)

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:

## Code review prechecks:

- [x] Code follows the style guidelines of this project
- [x] Code has been self-reviewed
- [x] Inline comments have been added for each method
- [x] I have made corresponding changes to the documentation
- [x] Code changes introduces no new problems or warnings
- [x] Test cases have been added 
- [x] Dependent changes have been merged and published in downstream modules

### DIFF
--- a/pallets/tasks/src/benchmarking.rs
+++ b/pallets/tasks/src/benchmarking.rs
@@ -144,34 +144,29 @@ benchmarks! {
 	}: _(RawOrigin::Root, task_id) verify { }
 
 
+
 	restart_batch {
-		let l in 1..1000;
 		create_shard::<T>(ETHEREUM);
 		let network = ETHEREUM;
-		let failed_batches: Vec<BatchId> = (0..l).map(|l| l as BatchId).collect();
-		let target_batch_id: u64 = (l - 1).into();
-
-		for batch_id in &failed_batches {
-			let task_id = Pallet::<T>::create_task(
-				network,
-				Task::SubmitGatewayMessage { batch_id: *batch_id }
-			);
-			BatchTaskId::<T>::insert(batch_id, task_id);
-			TaskNetwork::<T>::insert(task_id, network);
-		}
-
-		FailedBatchIds::<T>::put(failed_batches.clone());
-	}: _(RawOrigin::Root, target_batch_id) verify {
-		let new_task_id = TaskIdCounter::<T>::get();
+		let batch_id = BatchIdCounter::<T>::get();
+		let initial_task_id = Pallet::<T>::create_task(
+			network,
+			Task::SubmitGatewayMessage { batch_id }
+		);
+		FailedBatchIds::<T>::insert(batch_id, ());
+		BatchTaskId::<T>::insert(batch_id, initial_task_id);
+		TaskNetwork::<T>::insert(initial_task_id, network);
+	}: _(RawOrigin::Root, batch_id) verify {
+		let new_task_id = initial_task_id + 1;
 		assert_eq!(
-			BatchTaskId::<T>::get(target_batch_id),
+			BatchTaskId::<T>::get(batch_id),
 			Some(new_task_id),
 			"New task not created"
 		);
 		assert!(
-			!FailedBatchIds::<T>::get().contains(&target_batch_id),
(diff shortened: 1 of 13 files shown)


pallets/tasks/src/lib.rs
```diff
@@ -434,6 +449,23 @@ pub mod pallet {
...
+
+		#[pallet::call_index(14)]
+		#[pallet::weight(<T as Config>::WeightInfo::restart_batch())]
+		pub fn restart_batch(origin: OriginFor<T>, batch_id: BatchId) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+			let old_task_id = BatchTaskId::<T>::get(batch_id).ok_or(Error::<T>::InvalidBatchId)?;
+			let network = TaskNetwork::<T>::get(old_task_id).ok_or(Error::<T>::UnknownTask)?;
+			let new_task_id = Self::create_task(network, Task::SubmitGatewayMessage { batch_id });
+			BatchTaskId::<T>::insert(batch_id, new_task_id);
+			FailedBatchIds::<T>::mutate(|ids| {
```
libreamartists: Cost of execution increases proportional to the length of this storage item so ideally there would be a weight hint input which the extrinsic required to be equal to the length of this storage item which is used  when the weight is computed.

libreamartists: But this could be a follow up

pallets/tasks/src/benchmarking.rs
```diff
@@ -143,5 +143,29 @@ benchmarks! {
...
+
+	restart_batch {
+		create_shard::<T>(ETHEREUM);
+		let network = ETHEREUM;
+		let batch_id = BatchIdCounter::<T>::get();
+		let initial_task_id = Pallet::<T>::create_task(
+			network,
+			Task::SubmitGatewayMessage { batch_id }
+		);
+		FailedBatchIds::<T>::mutate(|ids| ids.push(batch_id));
```
libreamartists: benchmark assumes length 1 but will probably be much larger in practice. Even if we do the weight hint in a follow up, I would prefer to overestimate the weight than underestimate it like this. So might be better to include like 100 failed batches here or whatever a sensibly high number is.

peektism (approved):
Looks great! Only nit was the weight hint but that can be in a follow up. Benchmarking should probably assume a worse case though (conservatively large number of failed batches)

peektism (approved):

pallets/tasks/src/lib.rs
```diff
@@ -260,6 +265,10 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type BatchTaskId<T: Config> = StorageMap<_, Blake2_128Concat, BatchId, TaskId, OptionQuery>;
 
+	/// List of failed batches.
+	#[pallet::storage]
+	pub type FailedBatchIds<T: Config> = StorageValue<_, Vec<BatchId>, ValueQuery>;
```
dvc94ch:
```suggestion
	pub type FailedBatchIds<T: Config> = StorageMap<_, BatchId, (), ValueQuery>;
```

Using a HashSet is way more efficient than using a vec...

dvc94ch (approved):
looks good. we just need to avoid vecs especially if a storage map can be used...

pallets/tasks/src/lib.rs
dvc94ch: commented on that, the solution is to use a StorageMap instead of a StorageValue<Vec<_>>

PR #1557, opened 2025-02-17
Introduce end-to-end testing crate

agryaznov:
This adds crate to be used as our integration testing framework. 

Currently we have a `smoke` test which basically does the same as `tc-cli smoke-test 2 3` but in the form of a normal integration test. Just run:
```sh
cargo test --tests smoke
```

and it will 
1. Spawn the environment, starting all the needed containers. 
2. Run the test. 
3. Tear down the environment, stopping all the containers. 

Next thing to be done is to add GMP assets bridge e2e test.


config/envs/local/local-evm-e2e.yaml
```diff
@@ -0,0 +1,48 @@
...
+    batch_gas_limit: 10000000
+    gmp_margin: 0.0
+    shard_task_limit: 50
+    route_gas_limit: 10000000
+    route_base_fee: 1400000000
+    shard_size: 1
+    shard_threshold: 1
+chronicles: 
+  - http://localhost:8080
+  - http://localhost:8081
```
dvc94ch, line 48: what's the difference between this config and local-evm.yaml?

e2e-tests/tests/smoke.rs
```diff
@@ -0,0 +1,60 @@
...
+
+mod common;
+
+use common::TestEnv;
+use time_primitives::NetworkId;
+
+const SRC: NetworkId = 2;
+const DEST: NetworkId = 3;
+
+#[tokio::test]
```
dvc94ch: mmh, I like it. however how do we deal with concurrent tests? this seams to call docker compose up before the test which is a noop I guess if its already running, but then shuts down the containers when complete?

maybe add a test for the grpc backend too and show that they can run concurrently without issues and the approach works for me...

config/envs/local/local-evm-e2e.yaml
agryaznov: It's about setting localhost instead of docker host aliases, which won't work outside docker where the tests are being run

dvc94ch (approved):

e2e-tests/tests/smoke.rs
agryaznov: Yeah you're right, maybe at this stage it has to be run sequentially. Let's figure this out at the next stage when adding next test for the GMP brigde.

agryaznov:
@penumbra23 can we add Foundry to the GHA runner? We need `forge` for the [integration test](https://github.com/Analog-Labs/timechain/actions/runs/13388915169/job/37391904958?pr=1557#step:4:2917).

.github/workflows/pr-test-compose.yaml
```diff
@@ -7,6 +7,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 jobs:
+  if: false
```
agryaznov, line 10: @penumbra23 I disable this one because this PR introduces integration test which does the same

agryaznov:
Last left failing CI job has nothing to do with the changes introduced in this pr and has [been around for a while](https://github.com/Analog-Labs/timechain/actions/runs/13359269479/job/37306415107). It's (hopefully) going to be fixed with https://github.com/Analog-Labs/timechain/pull/1538

PR #1566, opened 2025-02-18
Move e2e tests from a dedicated crate to `tests/` at the workspace root

agryaznov:
A small enhancement regarding integration tests:

1. We declare a "nominal" _timechain_ package in the workspace `Cargo.toml`.
2. This allows us to get rid of a dedicated crate for e2e tests, and have them living in `tests/` directory, which is neat and in line with general convention for integration tests.


agryaznov:
@penumbra23 let's come up with the CI settings reliable enough to be accepted for all the future PRs

.github/workflows/pr-test-cargo-integration.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Build testsuite
+        uses: ./.github/actions/cargo-command
+        with:
+          command: test
+          args: --all-features --no-run
+      - name: Run testsuite
+        uses: ./.github/actions/cargo-command
+        with:
+          command: test
+          args: --all-features --tests
```
dvc94ch: can we make this `--all-features -p timechain`

.github/workflows/pr-test-cargo-unit.yaml
```diff
@@ -38,6 +38,6 @@ jobs:
         uses: ./.github/actions/cargo-command
         with:
           command: test
-          args: --all-features
+          args: --all-features --lib --bins
```
dvc94ch: and this `--all-features --exclude timechain`?

dvc94ch (approved):
don't forget to remove the e2e-tests folder...

agryaznov:
@penumbra23 could [port mappings](https://github.com/Analog-Labs/timechain/blob/860a7942722f6b9cf3de8834d52ffbf68b0500bc/docker-compose.yml#L117) break things for other ci jobs? If so we can rm it for now and disable the e2e flow until that's solved

haider-rs:
i guess both chain-2-evm and chain-3-evm is mapped to same port. If we change one to something else it shouldnt be an issue.

agryaznov:
> i guess both chain-2-evm and chain-3-evm is mapped to same port. If we change one to something else it shouldnt be an issue.

hmm to me it looks like they're mapped to [different](https://github.com/Analog-Labs/timechain/blob/860a7942722f6b9cf3de8834d52ffbf68b0500bc/docker-compose.yml#L117) ones: `8545` and `8546`
(otherwise it wouldn't work locally, but it works)

haider-rs:
correct my bad.

agryaznov:
For now let's merge this w the new e2e flow off, and without port mappings, until we figure out proper CI setting on the runners for that with @penumbra23 (tbd in a follow-up PR)

penumbra23:
One thing that can be done is export the port but avoid a fixed one:
```
ports:
- 9944
```

In this way the OS will assign a free one each time.

agryaznov:
> One thing that can be done is export the port but avoid a fixed one:
> 
> ```
> ports:
> - 9944
> ```
> 
> In this way the OS will assign a free one each time.

this particular one has been there for a [quite a while](https://github.com/Analog-Labs/timechain/pull/523/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R18) (was not touched in recent PRs) so I keep it as is here

FlorianFranzen:
@agryaznov This actually added an [invalid workflow file](https://github.com/Analog-Labs/timechain/actions/runs/13413226838) still failing CI in some status messages. Just though to let you know as it took me a while to even see where and why the CI was still failing. Nothing to do from your side, I will add a fix #1573.

PR #1567, opened 2025-02-19
[ci] Tune up cargo test workflow run conditions

agryaznov:
**upd**: for now we just don't run integration tests in `.github/workflows/pr-test-cargo.yaml`, and keep tc-cli smoke in `.github/workflows/pr-test-compose.yaml` to unblock all failing CIs

Tunning this up further is a subj to a follow-up PRs 
cc @penumbra23 @dvc94ch 

We [turned off](https://github.com/Analog-Labs/timechain/pull/1557#discussion_r1959898556) the e2e smoke test job, as the same test now runs along all other ones in [Check testsuite](https://github.com/Analog-Labs/timechain/blob/77d119d59885988dd4c8773fd9dc32a44389f24b/.github/workflows/pr-test-cargo.yaml#L1) job. 

This change make the latter job run on the same conditions as the removed one used to.

### DIFF
--- a/.github/workflows/pr-test-cargo.yaml
+++ b/.github/workflows/pr-test-cargo.yaml
@@ -1,23 +1,7 @@
 name: Check testsuite
 on:
   pull_request:
-    paths:
-      - '.github/actions/cargo-command/**'
-      - '.github/workflows/pr-test-cargo.yaml'
-      - 'chronicle/**'
-      - 'config/subxt/**'
-      - 'docs/**'
-      - 'node/**'
-      - 'pallets/**'
-      - 'primitives/**'
-      - 'runtime/**'
-      - 'tc-cli/**'
-      - 'tc-subxt/**'
-      - 'tss/**'
-      - 'utils/**'
-      - 'Cargo.toml'
-      - 'Cargo.lock'
-      - 'rust-toolchain.toml'
+    types: [labeled, unlabeled, opened, reopened, synchronize]
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-test-cargo.yaml
```diff
@@ -1,23 +1,7 @@
 name: Check testsuite
 on:
   pull_request:
```
penumbra23: Using `paths-ignore` I think we can skip the following:
- .github
- docs
- slack-bot
- scripts

The only one I'm not sure is `config`, since sometimes I change the Dockerfiles which touch the code. @dvc94ch wdyt?

agryaznov: > + scripts

no we cant' skip this, as e2e tests use build script

dvc94ch: we have a test in tc-cli to make sure all the configs are valid and also docker files or docker image build script can change

.github/workflows/pr-test-cargo.yaml
```diff
@@ -38,6 +38,6 @@ jobs:
         uses: ./.github/actions/cargo-command
         with:
           command: test
-          args: --all-features
+          args: --all-features --lib --bins
```
dvc94ch, line 41: wouldn't it be simpler to just #[ignore] e2e tests and if you want to run them do `cargo test --ignored`?

agryaznov: no, it wouldnt, why?

agryaznov: see https://github.com/rust-lang/cargo/issues/10491#issuecomment-1074078912

PR #1573, opened 2025-02-19
[ci] make ci green again

agryaznov:
recent PR broke merging ci jobs, let's fix it


docker-compose.yml
```diff
@@ -126,6 +128,8 @@ services:
     environment:
       RUST_LOG: 'tc_subxt=debug,chronicle=debug,tss=debug,gmp_evm=info'
       RUST_BACKTRACE: 1
+    ports:
+    - 8081:8080
```
dvc94ch: the error was port already being allocated, you can't have any forwarded ports. you need to remove the evm port forwarding...

dvc94ch (commented):
the reason is that multiple ci runs may be happening concurrently on the same machine, so no test/ci-run can use a fixed port number...

docker-compose.yml
agryaznov: evm port forwarding was removed along with that one. this is chronicle's, my guess was it resulted in failing job for the chronicle container build.. so I just [reverted](https://github.com/Analog-Labs/timechain/commit/a49bdaef2a63e1ff44c3e27a994ffe9e0e1961cd#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3) `docker_compose.yml` to the version where merging ci succeed

but yeah, the reason seems to be somewhere else

agryaznov:
## current issue 

seems to be introduced [here](https://github.com/Analog-Labs/timechain/commit/b6f1a10c104133596ae08d475b82eaab7d3234e3#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecR680)

`docker build -t chronicle -f ./config/docker/Dockerfile.chronicle-release --build-arg PROFILE=testnet --build-arg FEATURES=testnet,develop .`

```
416.6   error[E0583]: file not found for module `migrations`
416.6     --> /build/runtime/src/lib.rs:88:1
416.6      |
416.6   88 | pub mod migrations;
416.6      | ^^^^^^^^^^^^^^^^^^^
416.6      |
416.6      = help: to create the module `migrations`, create file "/build/runtime/src/migrations.rs" or "/build/runtime/src/migrations/mod.rs"
416.6      = note: if there is a `mod migrations` elsewhere in the crate already, import it with `use crate::...` instead
416.6 
416.6   error[E0433]: failed to resolve: could not find `fix_pool_members` in `migrations`
416.6      --> /build/runtime/src/lib.rs:680:31
416.6       |
416.6   680 | type Migrations = migrations::fix_pool_members::RemoveBuggyPoolMembersMigration<Runtime>;
416.6       |                               ^^^^^^^^^^^^^^^^ could not find `fix_pool_members` in `migrations`

```

@4meta5 would you please fix it? (for some reason that job wasn't run at the time of your merge, but it definitely was broken at that time)

cc @dvc94ch @penumbra23

peektism:
@FlorianFranzen is removing the migration to fix it, it is weird because it compiles locally for me and the files and declarations are there

dvc94ch (approved):

FlorianFranzen:
Removed any migrations that were no longer needed (including the one bracking chronicle image builds).

Should be ready to go. Will merge it once CI passes.

PR #1594, opened 2025-02-25
ci: add reset workflow (only development env)

### LINKED ISSUE #1518: automate resetting development env

penumbra23:
## Description

This PR adds a workflow which reset the development timechain (including chronicles, GRPC nodes and validators). The workflow is tested separately with a local cluster (except the `tc-cli deploy` step which requires it to be merged since it's a local action).

Closes #1518


dvc94ch (approved):
lgtm thx!

PR #1596, opened 2025-02-25
Batch members offline

### LINKED ISSUE #1304: batch_member_offline
Identified during #1297  

There are multiple places throughout our custom pallets where successive `member_offline` calls make multiple independent changes to the same storage items. Moreover, the original trigger for the first `member_offline` call is usually timeouts which are executed in batches of members.

This optimization introduces a batch member offline function to ensure that the storage items in each of our custom pallets are updated as few times as possible for the required changes.

peektism:
Closes #1304 

- [x] implementation
- [x] update tests + benches

Follow up:
- update weights once #1615 is closed and benchmark runs work in PRs


FlorianFranzen (approved):
Looks good and taught me more about how to use the `Entry` API.

PR #1600, opened 2025-02-26
tc-cli: skip messages if thread isn't specified

### LINKED ISSUE #1599: tc-cli sends Slack messages even when the thread_id isn't specified
**Description**

As seen from the Slack messages in the env channel, tc-cli spams the channels even when the thread_id isn't specified.

**Expected vs. Actual Behavior**

tc-cli should only reply if both arguments are set.

**Steps to Reproduce**

Run any tc-cli command with only the `SLACK_CHANNEL_ID` env set

penumbra23:
## Description

Closes #1599

### DIFF
--- a/tc-cli/src/slack.rs
+++ b/tc-cli/src/slack.rs
@@ -83,9 +83,9 @@ impl Slack {
 
 		// If thread isn't declared or empty, we should skip message sending
 		let thread = match std::env::var("SLACK_THREAD_TS") {
-			Ok(t) if t.is_empty()=> None,
+			Ok(t) if t.is_empty() => None,
 			Ok(t) => Some(SlackTs::new(t)),
-			Err(e) => return Err(e.into())
+			Err(e) => return Err(e.into()),
 		};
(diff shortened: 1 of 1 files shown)


tc-cli/src/slack.rs
```diff
@@ -80,8 +80,13 @@ impl Slack {
...
 		let channel = SlackChannelId::new(channel);
-		let thread = std::env::var("SLACK_THREAD_TS")?;
-		let thread = SlackTs::new(thread);
+
+		// If thread isn't declared or empty, we should skip message sending
+		let thread = match std::env::var("SLACK_THREAD_TS") {
+			Ok(t) if t.is_empty() => None,
+			Ok(t) => Some(SlackTs::new(t)),
+			Err(e) => return Err(e.into()),
+		};
```
dvc94ch: mmh, so the issue is SLACK_THREAD_TS having an invalid value

dvc94ch: wouldn't the oneliner `anyhow::ensure!(!thread.is_empty());` work?

dvc94ch (approved):

PR #1601, opened 2025-02-26
adding cctp integration test

### LINKED ISSUE #1574: cctp integration test

haider-rs:
- [x] Add cctp integration test
- [x] Take CCTP contracts list instead of single contract.


docker-compose.yml
```diff
@@ -79,6 +79,8 @@ services:
     - 'anvil -b 2 --steps-tracing --order fifo --base-fee 0'
     environment:
         ANVIL_IP_ADDR: '0.0.0.0'
+    ports:
+      - '8545:8545'
```
dvc94ch: doesn't work in ci...

dvc94ch: lets please avoid portforwarding

docker-compose.yml
```diff
@@ -94,9 +96,13 @@ services:
     - '--target-url=ws://chain-2-evm:8545'
     - '--network-id=2'
     - '--backend=evm'
+    - '--cctp-sender=000000000000000000000000ab5976445202ac58cdf7da9cb5797a70e6be1cdc'
```
dvc94ch: how is this generated?

gmp/evm/src/lib.rs
```diff
@@ -739,6 +800,58 @@ impl IConnectorAdmin for Connector {
...
+			srcNetwork: self.network_id,
+			source: contract.into(),
+			destNetwork: dest_network,
+			dest: a_addr(dest),
+			nonce,
+			gasLimit: gas_limit as _,
+			data: cctp_msg.clone().abi_encode().into(),
+		};
+		let call = sol::GmpTester::sendMessageCall { msg: msg.clone() };
+		let _ = self.evm_call(contract, call, gas_cost, None, None).await?;
```
dvc94ch: why can't you use tc.send_message() here?

gmp/evm/src/lib.rs
```diff
@@ -739,6 +800,58 @@ impl IConnectorAdmin for Connector {
...
+		let burn_hash: [u8; 32] = sha3::Keccak256::digest(&msg_data).into();
+		let response = self.get_cctp_attestation(burn_hash).await?;
+		let attestation =
+			response.attestation.ok_or(anyhow::anyhow!("Failed to get msg attestation"))?;
+		let attestation = attestation.strip_prefix("0x").unwrap_or(&attestation);
+		let attestation_bytes = hex::decode(attestation).unwrap();
+		cctp_msg.attestation = attestation_bytes.into();
+		msg.data = cctp_msg.abi_encode().into();
+		let message_id = Into::<GmpMessage>::into(msg).message_id();
+		Ok(message_id)
```
dvc94ch: message id shouldn't depend on the payload if backends are allowed to change it. since the changed message is tss signed, this isn't a security issue either

gmp/grpc/src/lib.rs
```diff
@@ -247,7 +247,12 @@ impl IConnectorAdmin for Connector {
 		Ok(())
 	}
 	/// Deploys a test contract.
-	async fn deploy_test(&self, gateway: Address, tester: &[u8]) -> Result<(Address, u64)> {
+	async fn deploy_test(
+		&self,
+		_additional_params: &[u8],
```
dvc94ch: ?

tc-cli/src/lib.rs
```diff
@@ -1034,6 +1040,47 @@ impl Tc {
...
+				hex::encode(msg_id),
+				dest_network,
+				self.format_address(Some(dest_network), dest_addr)?,
+				gas_limit,
+				self.format_balance(Some(src_network), gas_cost)?,
+			),
+		)
+		.await?;
+		Ok(msg_id)
+	}
```
dvc94ch: seems like this unnecessarily duplicates send_message

tc-cli/src/main.rs
```diff
@@ -535,3 +537,57 @@ async fn real_main() -> Result<()> {
...
 	Ok(())
 }
+
+async fn exec_smoke(
+	tc: Tc,
+	src: NetworkId,
+	src_addr: Address,
+	dest: NetworkId,
+	dest_addr: Address,
+	smoke_type: SmokeType,
```
dvc94ch:
```suggestion
	payload: Vec<u8>,
```

tc-cli/src/lib.rs
```diff
@@ -77,7 +77,11 @@ impl Tc {
 					url: network.url.clone(),
 					mnemonic: env.target_mnemonic.clone(),
 					cctp_sender: None,
-					cctp_attestation: None,
+					// TODO move to config?
+					// This params is for checking cctp attestation with tc-cli
+					cctp_attestation: Some(String::from(
+						"https://iris-api-sandbox.circle.com/attestations/",
+					)),
```
dvc94ch: guess this is probably an env variable like coinmarketcap?

docker-compose.yml
haider-rs: using universal factory `create2` call

docker-compose.yml
```diff
@@ -119,6 +121,8 @@ services:
     - '--target-url=ws://chain-3-evm:8545'
     - '--network-id=3'
     - '--backend=evm'
+    - '--cctp-sender=000000000000000000000000ab5976445202ac58cdf7da9cb5797a70e6be1cdc'
+    - '--cctp-attestation=https://iris-api-sandbox.circle.com/attestations/'
```
dvc94ch: this won't do anything right? cctp needs to be ignored or return an error on grpc backend...

gmp/evm/src/lib.rs
```diff
@@ -709,13 +741,22 @@ impl IConnectorAdmin for Connector {
...
-	async fn deploy_test(&self, gateway: Address, tester: &[u8]) -> Result<(Address, u64)> {
+	async fn deploy_test(
+		&self,
+		additional_params: &[u8],
+		gateway: Address,
+		tester: &[u8],
+	) -> Result<(Address, u64)> {
 		let bytecode = extract_bytecode(tester)?;
+		let cctp_contract = self.deploy_cctp_contract(additional_params, gateway, tester).await?;
+		tracing::info!("CCTP contract deployed at: {:?}", hex::encode(cctp_contract));
```
dvc94ch: there isn't a reason anymore to do this right? you can just:
1. deploy_test
2. subxt.set_cctp_address

pallets/networks/src/lib.rs
```diff
@@ -137,11 +139,32 @@ pub mod pallet {
...
-	/// Map storage for shard size.
+	/// Map storage for shard threshold.
 	#[pallet::storage]
 	pub type NetworkShardThreshold<T: Config> =
 		StorageMap<_, Blake2_128Concat, NetworkId, u16, OptionQuery>;
 
+	/// Map storage for cctp config.
+	#[pallet::storage]
+	pub type NetworkCctpConfig<T: Config> =
+		StorageMap<_, Blake2_128Concat, NetworkId, CctpConfig, OptionQuery>;
```
dvc94ch: what's cctpconfig?

tc-cli/src/config.rs
```diff
@@ -16,6 +19,7 @@ impl Config {
 		let config_path = path.join(config);
 		let config = std::fs::read_to_string(&config_path)
 			.with_context(|| format!("failed to read config file {}", config_path.display()))?;
+		tracing::info!("YAML config: {:?}", config);
```
dvc94ch: please no, this just spams the output. if we're going to keep it make it at most debug/trace

tc-cli/src/main.rs
```diff
@@ -430,36 +444,40 @@ async fn real_main() -> Result<()> {
...
+			src,
+			src_address,
+			dest,
+			dest_address,
+		} => {
+			tc.setup_test(src, dest).await?;
+			let src_addr = hex::decode(src_address)
+				.unwrap()
+				.try_into()
+				.expect("Unable to convert src_address to bytes32");
```
dvc94ch: address should be parsed with tc.parse_address

tc-cli/src/main.rs
```diff
@@ -430,36 +444,40 @@ async fn real_main() -> Result<()> {
...
+			let dest_addr = hex::decode(dest_address)
+				.unwrap()
+				.try_into()
+				.expect("Unable to convert dest_address to bytes32");
```
dvc94ch: what are these needed for anyway?

dvc94ch: can't you use the same addrs as returned by setup_test?

tc-cli/src/main.rs
```diff
@@ -430,36 +444,40 @@ async fn real_main() -> Result<()> {
...
-			let blocks = tc.read_events_blocks(exec).await?;
-			let msgs = tc.messages(dest, dest_addr, blocks).await?;
-			let msg = msgs
-				.into_iter()
-				.find(|msg| msg.message_id() == msg_id)
-				.context("failed to find message")?;
-			tc.print_table(None, "message", vec![msg]).await?;
-			tc.println(None, format!("received message after {} blocks", end - start))
-				.await?;
+			exec_smoke(tc, src, src_addr, dest, dest_addr, cctp_payload.encode()).await?;
```
dvc94ch: seems like there is still something missing. checking that the received message contains an attestation. so exec_smoke should return the received message, then you can check it here

dvc94ch (commented):
much better, liking it!

tc-cli/src/lib.rs
haider-rs: no this is outdated code. i have no idea why it was showing to you!

docker-compose.yml
haider-rs: yeah this is removed

gmp/evm/src/lib.rs
haider-rs: yup

primitives/src/network.rs
```diff
@@ -31,4 +40,44 @@ pub struct NetworkConfig {
...
+		Ok(CctpContracts(bounded_addresses))
+	}
+}
+
+#[cfg(feature = "std")]
+impl TryFrom<&str> for CctpUrl {
+	type Error = anyhow::Error;
+
+	fn try_from(s: &str) -> Result<Self> {
+		let url_bytes = s.as_bytes().to_vec();
```
libreamartists, line 88: shouldn't this use parity_scale_codec Encode to encode the String instead of conversion this way if we are using parity_scale_codec Decode to decode the bytes after it is in storage?

primitives/src/network.rs
```diff
@@ -1,17 +1,26 @@
...
 use scale_codec::{Decode, Encode};
+use scale_info::prelude::string::String;
+use scale_info::prelude::vec::Vec;
 use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
 
 pub const CHAIN_NAME_LEN: u32 = 50;
 pub const CHAIN_NET_LEN: u32 = 50;
+pub const MAX_CCTP_ADDRESSES: u32 = 50;
+pub const MAX_CCTP_URL_LEN: u32 = 200;
```
libreamartists, line 13: I might test this with a larger value to see if the truncation is what is causing that error that we talked about over the huddle

tc-cli/src/main.rs
```diff
@@ -430,36 +436,32 @@ async fn real_main() -> Result<()> {
...
-				if let Some(exec) = exec {
-					break (exec, end);
-				}
+			let _ = exec_smoke(tc, src, src_addr, dest, dest_addr, vec![]).await?;
+		},
+		Command::SmokeCctp { src, dest } => {
+			let (src_addr, dest_addr) = tc.setup_test(src, dest).await?;
+			tc.set_network_config(src, Some(src_addr)).await?;
+			tc.set_network_config(dest, Some(dest_addr)).await?;
+			let cctp_msg_data = "0000000000000000000000060000000000040CDD0000000000000000000000009F3B8679C73C2FEF8B59B4F3444D4E156FB70AA50000000000000000000000009F3B8679C73C2FEF8B59B4F3444D4E156FB70AA50000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001C7D4B196CB0C7B01D743FBC6116A902379C723800000000000000000000000033A2838EABD69A081CBEBE3F11DED4086C1CFC25000000000000000000000000000000000000000000000000000000000098968000000000000000000000000033A2838EABD69A081CBEBE3F11DED4086C1CFC25";
```
libreamartists, line 453: should this always be hardcoded? Ideally we would be able to test different cctp_msg_data that works right?

haider-rs: yeah this is kind of fetched from a already executed CCTP setup. since we dont have a CCTP setup contracts to integrate with circle we use it as a mock since the rest api for it passes.

primitives/src/network.rs
haider-rs: yeah thanks that might be the issue will check

peektism (approved):
benchmarking and weight changes lgtm

chronicle/src/tasks/mod.rs
```diff
@@ -193,11 +197,13 @@ impl TaskExecutor {
 			.get_gateway(network)
 			.await?
 			.context("no gateway registered")?;
+		let cctp_info = self.params.runtime.get_cctp_info(network).await?;
```
dvc94ch, line 200: there's a problem here. what if different chronicles read different cctp info because they're querying different blocks? we really should fix the block number in all storage queries... but is fine to do as a follow up.

chronicle/src/tasks/mod.rs
```diff
@@ -219,7 +225,17 @@ impl TaskExecutor {
 			let span2 = span.clone();
 			let handle = tokio::task::spawn(async move {
 				match exec
-					.execute(block_number, network, gateway, shard_id, task_id, task, span2)
+					// TODO add the cctp address here
```
dvc94ch: what's the TODO here?

primitives/src/gmp.rs
```diff
@@ -34,6 +36,141 @@ impl GmpParams {
...
+			mint_receipient,
+			burn_token,
+			nonce,
+			attestation,
+			message,
+			extra_data,
+		})
+	}
+}
+
```
dvc94ch: might be better to put all the encode/decode/cctp stuff in cctp.rs

dvc94ch (approved):
great work!

chronicle/src/tasks/mod.rs
haider-rs: oops outdated comment

agryaznov (approved):

PR #1607, opened 2025-02-26
[tc-cli] handle admin acc funding error better

### LINKED ISSUE #1604: tc-cli fails with 'no accounts found' message
**Description**

When running `./target/debug/tc-cli --env ./config/envs/development deploy` it fails after registering the networks with message:
`no accounts found`

The same behavior is detected locally and running on CI ([workflow run](https://github.com/Analog-Labs/timechain/actions/runs/13541738672/job/37854086600)).

**Steps to Reproduce**

`./target/debug/tc-cli --env ./config/envs/development deploy`

**Additional Information**

The message is omitted in the chain-connectors [HERE](https://github.com/Analog-Labs/chain-connectors/blob/f61444647199ab0fe970f7cff62dd30520b3c1e3/chains/ethereum/server/src/client.rs#L303)
most probably on a local Ethereum dev network running geth in dev mode (URL: wss://eth-local.development.analog.one)

agryaznov:
fixes https://github.com/Analog-Labs/timechain/issues/1604

currently `tc-cli deploy` fails when `!admin_funds.is_zero()` in config and the network doesn't support faucet

### DIFF
--- a/gmp/evm/src/lib.rs
+++ b/gmp/evm/src/lib.rs
@@ -405,7 +405,13 @@ impl IChain for Connector {
 	}
 	/// Uses a faucet to fund the account when possible.
 	async fn faucet(&self, balance: u128) -> Result<()> {
-		self.wallet.faucet(balance, None).await?;
+		match self.network_id() {
+			2 | 3 | 6 => {
+				let _ = self.wallet.faucet(balance, None).await?;
+			},
+			network_id => tracing::warn!("network {network_id} doesn't support faucet"),
+		}
+
 		Ok(())
 	}
 	/// Transfers an amount to an account.


dvc94ch (commented):
Think the right way to fix this is setting admin_funds: 0 for that network. Then the faucet isn't used...

dvc94ch:
@penumbra23 can you try setting the admin_funds to 0 for unsupported networks or when the account needs prefunding and see if that resolves the issue?

dvc94ch:
What I don't get is why we're trying to use faucet if network is prefunded or where the funds of the admin account disappeared

dvc94ch:
Basically what it's saying is you requested x amount of admin funds but your account is below that and we couldn't resolve the issue for you automatically in a not very user-friendly way

agryaznov:
well probably you're right;  this PR just patches it to work as [before](https://github.com/Analog-Labs/timechain/pull/1544/files#diff-1c4859bd538b138404e1f07cfc4c52407a3219cb772f9355685cc3434fcbc885L435) coz @penumbra23 said he's blocked with this issue.. if setting `admin_funds` solves it, then np I guess (not a big fan of hard-coding those network ids either)

dvc94ch:
Would still be great to have tc-cli provide a more descriptive error message...

dvc94ch:
will just open a new PR, thanks again @agryaznov !

PR #1613, opened 2025-02-27
Fix tc-subxt stream restart test

### LINKED ISSUE #1612: test_finalized_stream_error_and_recovery sometimes gets stuck
test in tc-subxt `finalized_stream_error_and_recovery` sometimes gets stuck and not pass

The Arc<Mutex<bool>> flag to set stream offline and online taskes some time for setup due to which there are chances for a block to get mixed add some wait for it to properly set and pass the test

haider-rs:
resolves:
https://github.com/Analog-Labs/timechain/issues/1612


dvc94ch (approved):
thx!

PR #1619, opened 2025-02-27
[tc-cli] add `dump-state` and `load-state` commands

agryaznov:
Might be helpful to store/load specific state of local evm chain (anvil) with:

```
tc-cli dump-state <NETWORK> [PATH]
```

```
tc-cli load-state <NETWORK> [PATH]
```

### DIFF
--- a/gmp/evm/src/lib.rs
+++ b/gmp/evm/src/lib.rs
@@ -830,6 +830,22 @@ impl IConnectorAdmin for Connector {
 		let stdout = std::str::from_utf8(&output.stdout)?;
 		Ok(stdout.into())
 	}
+	/// Dump anvil chain state
+	async fn dump_state(&self) -> Result<String> {
+		let output = Command::new("cast")
+			.arg("rp")
+			.arg("--rpc-url")
+			.arg(&self.url)
+			.arg("anvil_dumpState")
+			.output()
+			.context("failed to run cast")?;
+		if !output.status.success() {
+			let err = std::str::from_utf8(&output.stderr).ok().unwrap_or_default();
+			anyhow::bail!("cast exited with {}: {err}", output.status);
+		}
+		let stdout = std::str::from_utf8(&output.stdout)?;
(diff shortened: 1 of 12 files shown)


penumbra23:
@agryaznov Currently it's only for anvil? Any plans on taking snapshots from different chains? Would be good if this works with geth's node

tc-cli/src/main.rs
```diff
@@ -531,6 +539,18 @@ async fn real_main() -> Result<()> {
 		Command::RetryFailedBatch { batch_id } => {
 			tc.restart_failed_batch(batch_id).await?;
 		},
+		Command::DumpState { network, path } => {
+			let path = path.unwrap_or("anvil_state.txt".into());
```
dvc94ch, line 543: better name would probably be the network id. and we want to snapshot all networks into a directory including the chronicle keys and timechain

gmp/evm/src/lib.rs
```diff
@@ -830,6 +830,50 @@ impl IConnectorAdmin for Connector {
...
+			.json(&body)
+			.send()
+			.await?
+			.json()
+			.await?;
+
+		json["result"]
+			.as_str()
+			.map(|s| s.to_owned())
+			.ok_or(anyhow!("invalid rpc response"))
```
dvc94ch, line 852: does this work for any evm chain or just anvil?

agryaznov: just anvil, called rpc method is provided by anvil

tc-cli/src/main.rs
agryaznov: I thought that way as well, but then realized all other commands use `network` name for that arg...

agryaznov:
> @agryaznov Currently it's only for anvil? Any plans on taking snapshots from different chains? Would be good if this works with geth's node

anvil only

dvc94ch (approved):

PR #1620, opened 2025-02-28
ci: separate docker compose runs by run ID

penumbra23:
## Description

This PR changes the way logs are extracted and saved from the integration test steps. The prior solution used S3 buckets, which had the issue of not capturing logs before the `Upload logs` step (container didn't even start, early errors, etc.).

This solution has a prerequisite - a promtail instance needs to run on the self-hosted runner (we can always add the step in the CI if this doesn't perform well). The promtail config looks something like this:

```
server:
  log_level: info
clients:
  - url: LOKI_URL
positions:
  filename: /tmp/positions.yaml

scrape_configs:
  - job_name: docker
    docker_sd_configs:
      - host: unix:///var/run/docker.sock
        refresh_interval: 5s
    pipeline_stages:
      - static_labels:
          job: compose-test-logs
    relabel_configs:
      - source_labels: [__meta_docker_container_id]
        target_label: container_id
      - source_labels: [__meta_docker_container_name]
        target_label: container_name
      - source_labels: [__meta_docker_container_label_com_docker_compose_project]
        target_label: run_id
```

By running the docker compose services specifying the `project` (`-p` flag) with the value of  workflow run ID, promtail translates it to a `run_id` label for querying the exact run. It looks something like this:
image

The above image captures logs from containers running in this PR (run ID 13584257927).

LogQL query would be: `{job="compose-test-logs", run_id="13584257927"}`

### DIFF
--- a/.github/workflows/pr-test-compose.yaml
+++ b/.github/workflows/pr-test-compose.yaml
@@ -33,7 +33,7 @@ jobs:
         id: tc-cli
         run: |
           cat << EOF >> $GITHUB_OUTPUT
-          TC_CLI=docker compose run --rm --remove-orphans \
+          TC_CLI=docker compose -p $GITHUB_RUN_ID run --rm --remove-orphans \
             --env="TIMECHAIN_MNEMONIC=${{ secrets.TIMECHAIN_MNEMONIC }}" \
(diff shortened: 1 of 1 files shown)


dvc94ch (approved):
this might fix a lot of ci issues we have, great find!

PR #1624, opened 2025-02-28
envs: fix chronicle URLs to point to local DNS names

penumbra23:
## Description

This PR fixes the `waiting for chronicle` CI error. The chronicles URLs were FQDNs, including the network and replica. Since [THIS](https://github.com/Analog-Labs/timechain/pull/1620) PR adds project for every run, the FQDN are dynamic and change per run, so can't really use them like that anymore. Instead, now it's pointing to local hostnames (local in it's network is more correct), so it won't affect anything outside the network.

**NOTE**: Don't merge this PR until every CI passes, need to have a stable checkpoint

### DIFF
--- a/config/envs/local/local-grpc-tss.yaml
+++ b/config/envs/local/local-grpc-tss.yaml
@@ -48,9 +48,9 @@ networks:
     shard_size: 3
     shard_threshold: 2
 chronicles: 
-  - http://timechain-chronicle-0-grpc-1:8080
-  - http://timechain-chronicle-0-grpc-2:8080
-  - http://timechain-chronicle-0-grpc-3:8080
-  - http://timechain-chronicle-1-grpc-1:8080
-  - http://timechain-chronicle-1-grpc-2:8080
-  - http://timechain-chronicle-1-grpc-3:8080
+  - http://chronicle-0-grpc:8080
+  - http://chronicle-0-grpc:8080
+  - http://chronicle-0-grpc:8080
+  - http://chronicle-1-grpc:8080
+  - http://chronicle-1-grpc:8080
+  - http://chronicle-1-grpc:8080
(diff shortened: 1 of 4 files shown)


agryaznov (approved):
nice, thx!

dvc94ch:
just fyi this would break local tss testing by using the `docker compose --scale=chronicle-evm=3`

dvc94ch:
see here for a config to be used with --scale: https://github.com/Analog-Labs/timechain/blob/development/config/envs/local/local-grpc-tss.yaml

config/envs/local/local-grpc-tss.yaml
```diff
@@ -48,9 +48,9 @@ networks:
...
-  - http://timechain-chronicle-0-grpc-3:8080
-  - http://timechain-chronicle-1-grpc-1:8080
-  - http://timechain-chronicle-1-grpc-2:8080
-  - http://timechain-chronicle-1-grpc-3:8080
+  - http://chronicle-0-grpc:8080
+  - http://chronicle-0-grpc:8080
+  - http://chronicle-0-grpc:8080
+  - http://chronicle-1-grpc:8080
+  - http://chronicle-1-grpc:8080
+  - http://chronicle-1-grpc:8080
```
dvc94ch: I guess you already see the problem here...

penumbra23: Yeah, just figured out while I was reiterating, are we using service replicas for tss testing? What's the implication of actually running a different service?

penumbra23: It would actually resolve all IPs of each replica since it considers it to be the same service horizontally scaled

dvc94ch: Mainly a long docker compose file with lots of duplicate definitions. This way we can have a single docker compose file and scale to an arbitrary shard size

dvc94ch: Every chronicle has different keys, so this won't work. But it's fine for now, we'll figure something out

penumbra23: Reverted the changes in the tss config, we'll deal later with that (local testing shouldn't be affected)

PR #1625, opened 2025-03-01
remove payload from msg id

### LINKED ISSUE #1623: don't include payload in message id
> the reason for this is that with cctp we have the following flow:
>
> 1.user submits cctp message, message id is computed
> 2. gateway emits message created event
> 3. chronicle reads the event, sees its a cctp message and includes the attestation in the payload
> 4. shard threshold signs this using `hash(msg_id, hash(payload))`
> 5. then the rest is the same as regular gmp, submit, exec, etc
> 
> so we want chronicles to be able to modify the payload without changing the message id and we want the shard to threshold sign the modified message

haider-rs:
Fixes #1623 

Need to have a PR in analog-gmp for removing message_bytes from message_data and change in this branch
- [x] modify analog-gmp to remove GmpMessage bytes and update git reference here 
   done in https://github.com/Analog-Labs/analog-gmp/pull/46

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "analog-gmp"]
 	path = analog-gmp
 	url = git@github.com:Analog-Labs/analog-gmp.git
-	branch = fix-message-id
+	branch = main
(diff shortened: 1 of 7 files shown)


dvc94ch:
No changes to analog-gmp required? That's the hard part...

Issue #1615, opened 2025-02-27
runtime-benchmarks: fix develop  and add mainet and testnet


FlorianFranzen:
It seems the `develop` benchmark is not using the correct flags to build and we are still missing a runner to run mainnet and testnet benchmarks.

peektism:
Indeed, I noticed this when trying to run benchmarks for #1596

FlorianFranzen:
@4meta5 I will open a draft PR with the updated CI config in the next 24h, then we can work from there.

FlorianFranzen:
@4meta5 All benchmarks pass now, will cleanup the CI config today, but your work is done here.

PR #1628, opened 2025-03-02
Batch Shards Offline

### LINKED ISSUE #1342: batch_shard_offline
Shards are set offline in batches in terms of control flow but the storage items updated per shard set offline are not accessed in the most optimal way. The optimal way is to read/write once per batch instead of per shard.

peektism:
Closes #1342 

Follow up:
- run benchmarks to update weights (once #1615 is addressed first)

### DIFF
--- a/pallets/shards/src/lib.rs
+++ b/pallets/shards/src/lib.rs
@@ -31,10 +31,10 @@
 //! ### Force Shard Offline Flow
 //!
 //! The force shard offline process starts with ensuring the request is from a root user. Upon
-//! confirmation, the system calls the `remove_shard_offline` function to begin the shard removal
-//! process. This involves removing the shard state, retrieving the network details, and scheduling
-//! the `shard_offline` task. The process also includes draining and removing shard members,
-//! removing members from the `MemberShard`, and concludes with logging the `ShardOffline` event.
+//! confirmation, the system removes the shard. This involves removing the shard state,
+//! retrieving the network details, and scheduling the `shard_offline` task. The process also
+//! includes draining and removing shard members, removing members from the `MemberShard`, and
+//! concludes with logging the `ShardOffline` event.
 //!
 //! ### Ready Flow
 //!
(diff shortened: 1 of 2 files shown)


pallets/shards/src/lib.rs
```diff
@@ -428,25 +428,73 @@ pub mod pallet {
...
 			});
+
+			// Batch remove offline shards
+			if !shards_to_remove.is_empty() {
+				Self::batch_remove_shards_offline(shards_to_remove);
+			}
+
+			// Remove the timeout counter for this block
 			DkgTimeoutCounter::<T>::remove(n);
+
```
libreamartists:
```suggestion
```

pallets/shards/src/lib.rs
```diff
@@ -428,24 +428,71 @@ pub mod pallet {
...
 		pub(crate) fn timeout_dkgs(n: BlockNumberFor<T>) -> Weight {
 			let mut num_timeouts = 0u32;
-			// Iterate over DKG timeouts
+			let mut shards_to_remove = Vec::new();
+
+			// Collect all shard IDs that need to be removed
 			DkgTimeout::<T>::drain_prefix(n).for_each(|(shard_id, _)| {
 				if let Some(status) = ShardState::<T>::get(shard_id) {
 					if matches!(status, ShardStatus::Created | ShardStatus::Committed) {
-						Self::remove_shard_offline(shard_id);
```
dvc94ch, line 443: remove_shard_offline still used?

dvc94ch (approved):
looks good! you can write clean readable code! want that your PRs to always look like this...

pallets/shards/src/lib.rs
dvc94ch: I see, its used in force shard offline. would be good to remove the duplication

libreamartists: yeah it's only used there now so I can move it into that

dvc94ch: would be better to use the batched version in force shard offline, instead of having two versions doing basically the same thing...

libreamartists: I agree, done in https://github.com/Analog-Labs/timechain/pull/1628/commits/9215c3727c5c11871e53697ffc7c1703be37770e

PR #1630, opened 2025-03-03
Parallelize tc-cli deployment.

dvc94ch:
- deployment of networks
- registering routes
- deployment of chronicles
- funding of chronicles
- registering shards
- deploying testers


haider-rs (approved):
Thanks

PR #1635, opened 2025-03-04
ci: tag tc-cli image after runtime upgrade

### LINKED ISSUE #1561: deploying on an env should tag the images with the env

penumbra23:
## Description

This PR builds and pushes a tc-cli OCI image after a runtime upgrade. The image is tagged with the environment where the runtime upgrade was done (development, integration, etc.). In this way after a runtime upgrade has been done, the tc-cli latest environment image should be inline with the chain's metadata.

Closes #1561

### DIFF
--- a/.github/actions/build-contracts/action.yaml
+++ b/.github/actions/build-contracts/action.yaml
@@ -1,0 +1,47 @@
+name: "Install dependencies and build GMP contracts"
+description: "Action to install all tc-cli deps and build contracts"
+
+inputs:
+  version:
+    description: "Docker image version tag"
+    required: true
+    default: "dev"
+  token:
+    description: "Github API Token"
+    required: true
+  environment:
+    description: "Target env for the tc-cli dispatch"
+    required: true
+  args:
+    description: "Tc-cli arguments"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Cargo
+      uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: stable
+        target: x86_64-unknown-linux-musl,wasm32-unknown-unknown
+        components: rust-src
+    - name: Install musl deps
+      shell: bash
+      run: sudo apt-get update && sudo apt-get install -y musl-tools
+    - name: Rust cache
+      uses: Swatinem/rust-cache@v2
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+    - name: Build contracts
+      shell: bash
+      run: forge build --root analog-gmp --optimize --optimizer-runs=200000 --evm-version=shanghai --use=0.8.25 --force
+    # At the moment mainnet doesn't support GMP, so the only compatible metadata is the testnet one
+    - name: Build tc-cli
+      shell: bash
+      run: cargo build --profile testnet -p tc-cli --target x86_64-unknown-linux-musl --features testnet,develop
+    - name: Prepare context
+      shell: bash
+      run: |
+        mkdir docker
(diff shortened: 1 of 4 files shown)


.github/actions/build-contracts/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+    # At the moment mainnet doesn't support GMP, so the only compatible metadata is the testnet one
+    - name: Build tc-cli
+      shell: bash
+      run: cargo build --profile testnet -p tc-cli --target x86_64-unknown-linux-musl --features testnet,develop
+    - name: Prepare context
+      shell: bash
+      run: |
+        mkdir docker
+        cp target/x86_64-unknown-linux-musl/testnet/tc-cli docker/tc-cli
+        cp -rL config/envs docker/envs
```
dvc94ch: don't we already have some workflow to build tc-cli? seems like it doesn't copy the compiled contracts into the build context or the sources (which are used by `tc-cli debug-transaction`)

.github/workflows/merge-docker-tc-cli.yaml
```diff
@@ -44,28 +44,8 @@ jobs:
...
-      - name: Build contracts
-        run: forge build --root analog-gmp --optimize --optimizer-runs=200000 --evm-version=shanghai --use=0.8.25 --force
-      # At the moment mainnet doesn't support GMP, so the only compatible metadata is the testnet one
-      - name: Build tc-cli
-        run: cargo build --profile testnet -p tc-cli --target x86_64-unknown-linux-musl --features testnet,develop
-      - name: Prepare context
-        run: |
-          mkdir docker
-          cp target/x86_64-unknown-linux-musl/testnet/tc-cli docker/tc-cli
-          cp -rL config/envs docker/envs
```
dvc94ch, line 68: ups, maybe was my mistake and I didn't update the release dockerfile

dvc94ch: can you look into it?
- cast needs to be installed (think that already happens by untaring the forge release)
- analog-gmp needs to be copied to the tc-cli release image
- ANALOG_GMP_DIR needs to be set so the evm backend knows where to find the sources

penumbra23: Ok, all should be set now

.github/actions/build-contracts/action.yaml
penumbra23: Now it should be fixed, the build is done on the runner so all artifacts are copied to the context

dvc94ch (approved):
thx bran!

.github/workflows/merge-srtool-runtime.yaml
```diff
@@ -154,3 +153,16 @@ jobs:
...
+      uses: ./.github/actions/buildah-action
+      with:
+        image_name: ${{ env.TC_CLI_IMAGE }}
+        containerfile: ./config/docker/Dockerfile.tc-cli-release
+        context: docker
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+        registry: docker.io
+        push: 'true'
+        tags: ${{ github.event.inputs.environment }}
```
dvc94ch, line 168: one last thing, shouldn't we build this before performing the upgrade to ensure it fails first?

penumbra23: In that case it will overwrite the image used, first we need to use the current image, upgrade the runtime and build with the new runtime and tag it.
For example, current spec version is 18, do runtime upgrade 19 with tc-cli on 18 and rebuild and tag the image.

PR #1652, opened 2025-03-07
chronicle: add support for collecting traces

### LINKED ISSUE #1577: Tempo Setup
## Description

- Setting up Tempo locally in `docker-compose` or a separate `docker-compose-monitoring.yaml` (that we had for Grafana) so local testing can also happen
- Setting up Tempo on our cluster infrastructure

### LINKED ISSUE #1578: OpenTelemetry Chronicle integration
## Description

- Integrate opentelemetry inside the chronicle
- Hide all tracing support behind a `tracing` chronicle feature so the production binary doesn't bloat
- Inspect instrumentation calls and execution paths that need to be covered

penumbra23:
## Description

This PR adds the opentelemetry tracing support for the chronicle. They emit traces in the OTLP format making it compatible with any backend (Jaeger, Tempo, etc.).

The docker compose file spins up grafana and tempo with the `monitoring` profile. If the `TRACING_ENDPOINT` env isn't supplied, the traces aren't emitted.

Closes #1578 
Closes #1577 (with the addition of adding tempo to the supported infra charts https://github.com/Analog-Labs/infrastructure/pull/53)


chronicle/src/tasks/mod.rs
```diff
@@ -123,8 +129,13 @@ impl TaskParams {
...
 		task_id: TaskId,
 		task: Task,
-		span: Span,
 	) -> Result<()> {
+		let span = span!(
+			Level::INFO,
+			"executing_task",
+			task_id,
+			%task,
+		);
```
dvc94ch, line 136: why doesn't it have a parent? this means that it won't have the peer_id, block number, account id and other data when log filtering

chronicle/src/tasks/mod.rs
```diff
@@ -208,7 +220,6 @@ impl TaskExecutor {
 			}
 
 			let span = span!(
-				parent: span,
```
dvc94ch, line 217: also here...

chronicle/src/tasks/mod.rs
```diff
@@ -179,6 +190,7 @@ impl TaskExecutor {
 		}
 	}
 
+	#[tracing::instrument(skip(self))]
```
dvc94ch, line 194: what does this do?

chronicle/src/lib.rs
```diff
@@ -22,19 +31,51 @@ mod runtime;
 mod shards;
 mod tasks;
 
-pub fn init_logger() {
-	let filter = tracing_subscriber::EnvFilter::from_default_env()
-		.add_directive("chronicle=debug".parse().unwrap())
-		.add_directive("tss=debug".parse().unwrap());
-	tracing_subscriber::fmt()
```
dvc94ch, line 31: think this is also used by tests...

chronicle/src/tasks/mod.rs
penumbra23: This is actually creating a parent span named by the method and adding all arguments as attributes. See [here](https://docs.rs/tracing/0.1.41/tracing/index.html#spans-1)

chronicle/src/lib.rs
penumbra23: I guess it won't affect it much since it logs all debug stuff, how can I check if it removes anything?

dvc94ch: assumed it wouldn't build as tests use init_logger instead of init_opentelemetry. I guess if it builds its fine

chronicle/src/tasks/mod.rs
penumbra23: Added parent span

chronicle/src/lib.rs
penumbra23: Ok, tests were failing so I used init_opentelemetry in the chronicle tests

chronicle/src/tasks/mod.rs
dvc94ch: not needed?

dvc94ch (approved):

PR #1654, opened 2025-03-07
e2e tests crate

agryaznov:


dvc94ch:
if build is green, lgtm. one issue is why the lockfile grew so much?

agryaznov:
> if build is green, lgtm. one issue is why the lockfile grew so much?

oops, that's deps for the new tests, not adding them rn; removed.

dvc94ch (approved):

PR #1676, opened 2025-03-12
Don't fund chronicles.

dvc94ch:


haider-rs (approved):

dvc94ch:
doesn't work, chronicles need an existential deposit

Issue #1662, opened 2025-03-11
shard creation is slow and requires multiple restarts


dvc94ch:
1. reset the previous development env (completely wiped everything)
2. brought up the chronicles and did tc-cli deploy
3. shards were created, but timeout out committing for 5-10 minutes
4. one by one shards got online, but some chronicles panicked in the process with this message:
```
2025-03-10T15:31:37.112909Z ERROR tracing_panic: A panic occurred, panic.payload: "called `Result::unwrap()` on an `Err` value: 0x7fe043001a11210e25d8063908d58fb6d8ba6f5450c4c9ba6da6e0aaeda60d5b extrinsic failed with code: [3, 34, 8, 0, 0, 0, 98, 126, 125, 128, 233, 179, 0, 0], pallet idx: 0, variant idx: 1", panic.location: "/build/chronicle/src/shards/service.rs:319:51", panic.backtrace: disabled backtrace, panic.note: "run with RUST_BACKTRACE=1 environment variable to display a backtrace"
    at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-panic-0.1.2/src/lib.rs:82
```

peektism:
David added error messages so we will wait until this error is seen again before attempting a solution. #1680 implemented exponential backoff and also returned early for a few error conditions. The code therein may be reused once we have the error message returned from this issue in the future.

dvc94ch:
issue disappeared. redeployment on integration had no issues

PR #1680, opened 2025-03-13
Shard commitment exponential backoff

peektism:
WIP Addresses #1662 

- [x] implements exponential backoff for retrying shard commitment
- [x] checks shard status and online member status before shard commitment


dvc94ch:
don't think this is the right approach. we can probably close this. thanks for the effort!

PR #1683, opened 2025-03-13
Fix assertions.

dvc94ch:


pallets/tasks/src/lib.rs
```diff
@@ -229,6 +229,16 @@ pub mod pallet {
 	pub type TaskNetwork<T: Config> =
 		StorageMap<_, Blake2_128Concat, TaskId, NetworkId, OptionQuery>;
 
+	/// Map storage for register shard batches.
+	#[pallet::storage]
+	pub type ShardRegisterBatchId<T: Config> =
+		StorageMap<_, Blake2_128Concat, TssPublicKey, BatchId, OptionQuery>;
+
+	/// Map storage for register shard batches.
```
agryaznov:
```suggestion
	/// Map storage for unregister shard batches.
```

tc-cli/src/lib.rs
```diff
@@ -535,6 +536,14 @@ impl Tc {
 				registered = registered_shards.get(&network).unwrap().contains(&key);
 			}
 			let assigned = self.runtime.assigned_tasks(shard).await?.len();
+			let (batch_register, batch_unregister) = if let Some(key) = key {
```
agryaznov: can't this be put into the same if let block 2 lines above?

agryaznov (approved):
This probably requires benchmarks re-run, as weight for e.g. `prepare_batches()` might be affected

PR #1687, opened 2025-03-14
ci: tc-cli uses correct env version

penumbra23:
## Description

This PR forces tc-cli to use the correct image tag based on what environment it's being triggered


haider-rs (approved):
LGTM thx!

PR #1689, opened 2025-03-14
runtime: cleanup before testnet reset

FlorianFranzen:
This prepares the runtime code for the testnet reset:

- Add missing keys
- Allow individual council member governance on testnet
- Remove treasury pallet (and replace with virtual account for fees)
- Remove airdrop pallet from testnet
- Remove safe-mode pallet


dvc94ch (approved):

runtime/src/configs/governance.rs
```diff
@@ -59,32 +51,31 @@ pub type TechnicalSuperMajority =
 pub type TechnicalUnanimity =
 	pallet_collective::EnsureProportionAtLeast<AccountId, TechnicalCollective, 1, 1>;
 
+// Combine with root origin to allow easier benchmarking
 pub type EnsureRootOrTechnicalMember = EitherOfDiverse<EnsureRoot<AccountId>, TechnicalMember>;
 pub type EnsureRootOrHalfTechnical = EitherOfDiverse<EnsureRoot<AccountId>, TechnicalHalf>;
 
+/// Default admin origin on mainnet
+#[cfg(not(any(feature = "testnet", feature = "develop")))]
+pub type DefaultAdminOrigin = EnsureRootOrHalfTechnical;
```
libreamartists, line 60: I preferred the previous naming, this new naming requires looking up this definition

peektism (approved):

runtime/src/configs/governance.rs
FlorianFranzen: The problem is that  it need to be different depending on the environment and having different names means we need to feature gate every place this is used.

PR #1690, opened 2025-03-14
node: select default chain-spec based features

FlorianFranzen:
Minor comfort feature to run the right chain-spec by default.


dvc94ch (approved):
looks like you need to configure your editor to run `cargo fmt` before save...

FlorianFranzen:
I did but it seems my editor uses some different formatting. Weird, got to investigate what it actually runs.

PR #1701, opened 2025-03-17
Runtime storage hash

### LINKED ISSUE #1649: all substrate storage queries and runtime api calls should use a specific block

haider-rs:
closes: https://github.com/Analog-Labs/timechain/issues/1649

### DIFF
--- a/chronicle/src/lib.rs
+++ b/chronicle/src/lib.rs
@@ -186,7 +186,7 @@ pub async fn run_chronicle(
 		.await?;
 	loop {
 		let Some((hash, _)) = ticker.next().await else { continue };
-		if substrate.is_registered(hash).await? {
+		if substrate.is_registered(hash.into()).await? {
 			break;
 		}
 		tracing::warn!(parent: &span, "chronicle isn't registered");
@@ -214,7 +214,6 @@ mod tests {
 	use polkadot_sdk::sp_runtime::BoundedVec;
 	use scale_codec::Encode;
 	use std::time::Duration;
-	use tc_subxt::SubxtBlock;
 	use time_primitives::traits::IdentifyAccount;
 	use time_primitives::{AccountId, BlockHash, ChainName, ChainNetwork, ShardStatus, Task};
 
@@ -278,7 +277,7 @@ mod tests {
 		init_opentelemetry();
 
 		let mock = Mock::default().instance(42);
-		let block: SubxtBlock = BlockHash::from([0u8; 32]).into();
+		let block: BlockHash = BlockHash::from([0u8; 32]).into();
 		let network_id = mock.create_network(
 			ChainName(BoundedVec::truncate_from("rust".encode())),
 			ChainNetwork(BoundedVec::truncate_from("rust".encode())),
@@ -311,7 +310,7 @@ mod tests {
 		// Wait for the shard to be online.
 		loop {
 			tracing::info!("waiting for shard");
-			if mock.get_shard_status(shard_id).await.unwrap() != ShardStatus::Online {
+			if mock.get_shard_status(shard_id, block).await.unwrap() != ShardStatus::Online {
 				tokio::time::sleep(Duration::from_secs(1)).await;
 				continue;
 			}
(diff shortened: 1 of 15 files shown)


chronicle/src/lib.rs
```diff
@@ -129,7 +129,7 @@ pub async fn run_chronicle(
 	let mut ticker = substrate.finality_notification_stream();
 	// Initialize connector
 	let (chain, subchain) = loop {
-		let network = substrate.get_network(config.network_id).await?;
+		let network = substrate.get_network(config.network_id, None).await?;
```
dvc94ch: why use none when you have the block numbers from the ticker?

tc-cli/src/lib.rs
```diff
@@ -646,27 +649,27 @@ impl Tc {
...
-		Ok(self.runtime.message_executed_task(message).await?.is_some())
+		Ok(self.runtime.message_executed_task(message, None).await?.is_some())
 	}
 
 	pub async fn is_batch_executed(&self, batch: BatchId) -> Result<bool> {
-		Ok(self.runtime.batch_tx_hash(batch).await?.is_some())
+		Ok(self.runtime.batch_tx_hash(batch, None).await?.is_some())
 	}
 
 	pub async fn message_trace(
```
dvc94ch, line 736: think most of these should also have a block they're querying

tc-cli/src/main.rs
```diff
@@ -436,38 +437,43 @@ async fn real_main() -> Result<()> {
 			tc.println(None, hex::encode(msg_id)).await?;
 		},
 		Command::SmokeTest { src, dest } => {
+			let mut stream = tc.finality_notification_stream();
 			let testers = tc.setup_test().await?;
 
 			// collect shard batches
 			let mut batches = HashSet::new();
-			for shard in tc.shards().await? {
+			let (block_hash, _) = tc.latest_block().await?;
```
haider-rs, line 445: still thinks that the `Arc<Mutex<BlockHash>>` in `Tc` is better to avoid these calls and extending smoke-test duration. @dvc94ch what are your thoughts?

dvc94ch: Seems fine to me, will take a closer look probably Friday

e2e-tests/tests/smoke.rs
```diff
@@ -12,26 +12,27 @@ const DEST: NetworkId = 3;
...
+		let trace = tc.message_trace(SRC, msg_id, hash).await?;
 		let exec = trace.exec.as_ref().map(|t| t.task);
 		tracing::info!(target: "smoke_test", "waiting for message {}", hex::encode(msg_id));
 		id = Some(tc.print_table(id, "message", vec![trace]).await?);
 		if let Some(exec) = exec {
 			break (exec, end);
 		}
 	};
-	let blocks = tc.read_events_blocks(exec).await?;
+	let block = tc.latest_block().await?;
```
dvc94ch: you already have the block hash from the loop...

tc-cli/src/lib.rs
```diff
@@ -49,6 +49,7 @@ pub struct Tc {
 	runtime: SubxtClient,
 	connectors: HashMap<NetworkId, Arc<dyn IConnectorAdmin>>,
 	msg: Sender,
+	pub init_block: BlockHash,
```
dvc94ch:
```suggestion
```

tc-cli/src/main.rs
```diff
@@ -244,7 +244,7 @@ async fn real_main() -> Result<()> {
 	match args.cmd {
 		// balances
 		Command::Faucet { network } => {
-			tc.faucet(network).await?;
+			tc.faucet(network, tc.init_block).await?;
```
dvc94ch:
```suggestion
                        let (hash, _) = tc.latest_block().await?;
			tc.faucet(network, hash).await?;
```

dvc94ch (approved):
nice work! thanks @haider-rs for doing this incredibly boring task. much appreciated.

PR #1711, opened 2025-03-19
[GMP] use alloy instead of chain-connectors

agryaznov:


chronicle/src/shards/service.rs
```diff
@@ -365,7 +365,8 @@ where
 		self.outgoing_requests.push(Box::pin(poll_fn(|_| Poll::Pending)));
 
 		let task_params = self.task_params.clone();
-		let mut block_stream = task_params.block_stream().fuse();
+		let mut block_stream =
+			task_params.block_stream().await.expect("Failed to get block stream").fuse();
```
dvc94ch: the issue with this is what happens when the connection is lost? we don't resubscribe to the block_stream

chronicle/src/tasks/mod.rs
```diff
@@ -34,8 +34,8 @@ impl TaskParams {
 		self.connector.network_id()
 	}
 
-	pub fn block_stream(&self) -> Pin<Box<dyn Stream<Item = u64> + Send + '_>> {
-		self.connector.block_stream()
+	pub async fn block_stream(&self) -> Result<Pin<Box<dyn Stream<Item = u64> + Send + '_>>> {
+		self.connector.block_stream().await
```
dvc94ch: I guess we can move the retry logic to the chronicle instead of having it in the backend, but it needs implementing. the connectors used to take care of it

primitives/src/gmp.rs
```diff
@@ -6,8 +6,8 @@ use scale_info::{prelude::vec::Vec, TypeInfo};
 use serde::{Deserialize, Serialize};
 use sha3::{Digest, Keccak256};
 
-pub type Address = [u8; 32];
-pub type Gateway = Address;
+pub type Address32 = [u8; 32];
+pub type Gateway = Address32;
```
dvc94ch: lets remove the Gateway alias while we're at it
```suggestion
```

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
 pub struct Connector {
 	network_id: NetworkId,
-	wallet: Arc<Wallet>,
-	backend: Adapter<DefaultClient>,
+	rpc: Arc<CProvider>,
 	url: String,
+	signer: Arc<LocalSigner<SigningKey>>,
 	cctp_queue: Arc<Mutex<Vec<CctpRequest>>>,
 	// Temporary fix to avoid nonce overlap
 	wallet_guard: Arc<Mutex<()>>,
```
dvc94ch, line 82: don't think we need this anymore either if alloy supports concurrent transactions

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
 	/// Human readable connector account identifier.
-	fn address(&self) -> Address {
-		self.parse_address(&self.wallet.account().address).unwrap()
+	fn address(&self) -> Address32 {
+		t_addr(self.signer.address())
 	}
 	fn currency(&self) -> (u32, &str) {
-		let config = self.wallet.config();
-		(config.currency_decimals, config.currency_symbol)
+		(18, "ETH")
```
dvc94ch: this depends on the network, we shouldn't hard code it.

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
 	}
 	/// Uses a faucet to fund the account when possible.
-	async fn faucet(&self, balance: u128) -> Result<()> {
-		self.wallet.faucet(balance, None).await?;
-		Ok(())
+	async fn faucet(&self, _balance: u128) -> Result<()> {
+		Err(anyhow!("Faucet not supported"))
```
dvc94ch: need to implement this at least for anvil

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
-		self.wallet
-			.transfer(&AccountIdentifier::new(address), amount, None, None)
-			.await?;
+	async fn transfer(&self, address: Address32, amount: u128) -> Result<()> {
+		let to = a_addr(address);
+		let tx = TransactionRequest::default()
+			.with_from(self.signer.address())
+			.with_to(to)
+			.with_value(U256::from(amount));
+		let _tx_hash = self.rpc.send_transaction(tx).await?.watch().await?;
```
dvc94ch: would be good to log all tx hashes, although I know we aren't doing it consistently

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
+			.get_block(BlockId::finalized())
+			.await?
+			.map(|b| b.header.number)
+			.ok_or(anyhow!("failed querying finalized block"))
 	}
 	/// Stream of finalized block indexes.
-	fn block_stream(&self) -> Pin<Box<dyn Stream<Item = u64> + Send>> {
-		self.wallet.block_stream()
+	async fn block_stream(&self) -> Result<Pin<Box<dyn Stream<Item = u64> + Send>>> {
+		let subscription = self.rpc.subscribe_blocks().await?;
```
dvc94ch: are these finalized blocks?

gmp/evm/src/lib.rs
```diff
@@ -545,129 +270,159 @@ impl IConnector for Connector {
...
-		})?;
+
+		let address = a_addr(gateway);
+		let gw = IExecutorInstance::new(address, self.rpc.clone());
+
+		let _pending_tx =
+			gw.batchExecute(signature, message).gas(gas_limit).send().await.map_err(|err| {
+				tracing::info!("failed to submit batch: {:?}", err);
+				err.to_string()
+			})?;
```
dvc94ch: same here, we should log the tx hash. any idea what happens during chain reorgs? are these txs included in a block? or included in a finalized block?

dvc94ch (commented):
great work!

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
 		Self: Sized,
 	{
-		let (blockchain, private_key) = if params.blockchain == "anvil" {
-			let private_key = hex_literal::hex![
-				"ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
-			];
-			(Blockchain::Ethereum, Some(private_key))
-		} else {
-			(params.blockchain.parse()?, None)
-		};
```
dvc94ch: so we only need params.blockchain params.network to detect if we're running on anvil? if there's a better way of doing that, we should remove these from the ConnectorParams

gmp/evm/src/lib.rs
```diff
@@ -738,85 +487,98 @@ impl IConnectorAdmin for Connector {
 			data: payload.into(),
 		};
 		tracing::debug!("Sending GMP message: {:#?}", &msg);
+		// TODO why this contract is used here?
```
dvc94ch: send_message is only used for testing with the GmpProxy contract. Its so we can build a backend agnostic test suite

gmp/evm/src/lib.rs
```diff
@@ -882,12 +644,268 @@ impl IConnectorAdmin for Connector {
...
+		&self,
+		config: &DeploymentConfig,
+		call: Vec<u8>,
+	) -> Result<(Address20, u64)> {
+		let factory_address = a_addr(self.parse_address(&config.factory_address)?);
+
+		let tx = TransactionRequest::default()
+			.with_to(factory_address)
+			.with_chain_id(self.rpc.get_chain_id().await?)
+			// TODO why magic value
```
dvc94ch, line 740: no idea, @haider-rs where does this value come from?

gmp/evm/src/lib.rs
```diff
@@ -882,12 +644,268 @@ impl IConnectorAdmin for Connector {
...
+			.with_context(|| format!("tx {tx_hash} logs not found"))?;
+
+		let topic =
+			log.topics().first().with_context(|| format!("tx {tx_hash} topic not found"))?;
+
+		let contract_address = Address20::from_slice(&topic[12..]);
+		Ok((contract_address, receipt.block_number.unwrap()))
+	}
+
+	// TODO this needs refactoring: why deployer and contract bytecode are hard-coded?
```
dvc94ch, line 766: @haider-rs ?

gmp/evm/src/lib.rs
dvc94ch: its crucial that this returns an error or keeps trying until its done. if this fails to submit the batch then the timechain has messages that will never arrive. it needs to at least return an error in that case so that the failed batches can be monitored and resubmitted. if it returns Ok and then something fails later, that's the worst case scenario

gmp/evm/src/lib.rs
```diff
@@ -1,432 +1,160 @@
...
 
-fn t_addr(address: alloy_primitives::Address) -> Address {
-	let mut addr = [0; 32];
-	addr[12..32].copy_from_slice(&address.0[..]);
-	addr
+fn t_addr(address: Address20) -> Address32 {
+	address.into_word().into()
 }
 
+type CProvider = FillProvider<
```
penumbra23, line 61: Sorry to interfere here, but you can also use the `RootProvider` type instead of the generic one, since then we need to use the exact fillers that the type describes, right?

gmp/evm/src/lib.rs
haider-rs: This is universal factory deployment. And the tx was provided by Lohann to be deployed. Proxy deployment depends on it. and the tx signer is the address we fund actually.

gmp/evm/src/lib.rs
haider-rs: Deployment with factory was sometime failing with errors. Dont remember the error code yet. But this value was suggested by Lohann to be placed in order for the deployment with universal factory to succeed.

gmp/evm/src/lib.rs
agryaznov: This one has specific fillers we want, it's not generic. RootProvider alone won't fill anything for us

gmp/evm/src/lib.rs
agryaznov: alloy's nonce management didn't help with this, so keeping this guard here for now

gmp/evm/src/lib.rs
agryaznov: now they are, ported `FinalizedBlockStream` from chain-connectors

chronicle/src/shards/service.rs
agryaznov: With `FinalizedBlockStream`, this `IChain::block_stream()` is infallible again. 

alloy-pubsub service [handles](https://github.com/alloy-rs/alloy/blob/561ae358fcbc1ced3713b8074a408cae67aae9d2/crates/pubsub/src/service.rs#L208) re-connections for us

docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
     - '--node-key=0000000000000000000000000000000000000000000000000000000000000001'
     - '-ltxpool=trace,basic_authorship=trace,runtime=trace'
     ports:
-      - 9944
+      - 9944:9944
```
dvc94ch: not again, thought we agreed not to do this (once we use the e2e tests it doesn't matter anymore, then I can accept this)

docker-compose.yml
```diff
@@ -143,7 +143,7 @@ services:
     image: 'analoglabs/tc-cli-develop'
     environment:
       RUST_BACKTRACE: 1
-      RUST_LOG: 'gmp_evm=info'
+      RUST_LOG: 'gmp-evm=info'
```
dvc94ch: are you sure? usually its the module name not the crate name that needs logs enabled

gmp/grpc/src/lib.rs
```diff
@@ -4,7 +4,7 @@ use std::ops::Range;
 use std::pin::Pin;
 use std::sync::Arc;
 use time_primitives::{
-	Address32, BatchId, ConnectorParams, Gateway, GatewayMessage, GmpEvent, GmpMessage, IChain,
+	Address32 as Address, BatchId, ConnectorParams, GatewayMessage, GmpEvent, GmpMessage, IChain,
```
dvc94ch: why did you alias it and revert again?

dvc94ch:
released a new peernet version

dvc94ch (commented):
so no removing blockchain/network from ConnectorParams and moving tc to use a single string?

dvc94ch (approved):
LGTM! lets merge this

agryaznov:
> so no removing blockchain/network from ConnectorParams and moving tc to use a single string?

let's do that in a follow-up PR

PR #1734, opened 2025-03-24
Limit feeless heartbeat per timeout period

### LINKED ISSUE #1673: Limit feeless heartbeat submissions per timeout period
We may want to add back the error committed in https://github.com/Analog-Labs/timechain/pull/1670/commits/4430e36296358133b3a7ac8c411add11c1ccadc6 to limit feeless heartbeat submissions per timeout period

This error was removed from #1670 so that the PR does not create unforeseen problems, but it should be added back.

peektism:
Closes #1673 

- [x] member pallet changes
- [x] chronicle changes

Follow up: heartbeat E2E integration tests extracted in #1739 to not hold up this PR


pallets/members/src/lib.rs
```diff
@@ -247,6 +249,7 @@ pub mod pallet {
 			Ok(())
 		}
 		fn execute_send_heartbeat(member: AccountId) -> DispatchResult {
+			ensure!(Heartbeat::<T>::get(&member).is_none(), Error::<T>::AlreadySubmittedHeartbeat);
```
agryaznov, line 252:
```suggestion
			ensure!(!Heartbeat::<T>::contains_key(&member), Error::<T>::AlreadySubmittedHeartbeat);
```

libreamartists: Is one better than the other? Please show me if so because I was under the impression that both take up a read

libreamartists: interesting, read that contains_key is more efficient because it does not decode the value but instead just checks inclusion. The value is `()` so decoding cost is nothing in this case but still a good suggestion.

dvc94ch (approved):
lgtm, thx!

PR #1759, opened 2025-03-29
polkadot-sdk: v1.16.6 -> v1.17.3

FlorianFranzen:
This updates the polkadot-sdk to v1.17 which includes various improvements we would like to use. 

Additional changes:
 - Remove bridge pallet from runtime
 - Expose new upstream benchmarks
 - Add separate benchmark for feeless tx extension
 - various cleanups, dead-code removal


FlorianFranzen:
This has been deployed to staging (both node and runtime) and is currently monitored there.

peektism (approved):

pallets/members/src/benchmarking.rs
```diff
@@ -62,5 +63,25 @@ benchmarks! {
 		}
 	}
 
+	is_member {
```
libreamartists, line 66: where is this benchmark used?

runtime/src/configs/custom.rs
```diff
@@ -179,29 +149,69 @@ where
...
+{
+	const IDENTIFIER: &'static str = "PrevalidateFeeless";
+	type Implicit = ();
+	type Pre = ();
+	type Val = ();
+
+	fn weight(&self, call: &T::RuntimeCall) -> Weight {
+		if Self::is_feeless(call) {
+			// TODO: Use dynamic/cached member count?
+			return <T as pallet_members::Config>::WeightInfo::is_member(200);
```
libreamartists, line 184: do we want to make a follow up issue for caching this count in storage? I think it is fine to leave this as is for this PR

pallets/members/src/benchmarking.rs
libreamartists: I found it in the PrevalidateFeeless impl, interesting

runtime/src/configs/custom.rs
```diff
@@ -179,29 +149,69 @@ where
...
+		}
+
+		Weight::zero()
+	}
```
FlorianFranzen, line 188: @4meta5 The benchmark is used to estimate the overhead of checking membership status of feeless transactions.

runtime/src/configs/custom.rs
FlorianFranzen: I would assume it is worthwhile to know the number of chronicles, so if we could add this to the members pallet, we could use that here.

pallets/members/src/benchmarking.rs
```diff
@@ -62,5 +63,25 @@ benchmarks! {
...
+		// Pick different member to query each benchmark
+		let c in 0..(NUM_MEMBERS-1).into();
```
libreamartists, line 68: We store members as keys in a map so the cost of `is_member_registered` does not depend on the number of members in the map unlike if we stored them in a vec or btree_map

libreamartists: I do not think this complexity parameter is necessary or valuable with this in mind

runtime/src/configs/custom.rs
libreamartists: I actually think we should remove the complexity parameter after reviewing the benchmarking code more closely. I do not see why the cost would depend on the number of members in the map at all.

pallets/members/src/benchmarking.rs
FlorianFranzen: True, weights seem to indicate that too. Will fix this in the follow-up benchmark commit.

PR #1765, opened 2025-04-01
Patch finish task panic

### LINKED ISSUE #1763: Fix tasks pallet finish_task panic
There is a panic in finish_task specifically when a BatchExecuted task is finished due to a submit result call which can lead to finish_task being called twice.

1. reproduce issue in unit test
2. patch bug

peektism:
Closes #1763 

The panic is an edge case which happens when the read task ID is equal to the batch task ID. This causes finish_task to be called twice as a consequence of a single call to `submit_task_result`.

Fixes:
- [x] remove panic from calling finish_task twice as consequence of a single submit_task_result call
- [x] keep the first result submitted to finish_task and ignore the second call to it for the edge case described (because the first result is the consequence of processing the event while the second is a user-submitted result)

Tests:
- [x] unit test showing that TaskShard is cleared as a result of finish_task (unchanged yet expected behavior)
- [x] regression test showing that submit_task_result panics in the edge case described at the top of this description
- [x] regression test showing that the first result is kept in storage and only its associated event is emitted

Follow up discussion/issue: https://github.com/Analog-Labs/timechain/issues/1766

### DIFF
--- a/pallets/tasks/src/lib.rs
+++ b/pallets/tasks/src/lib.rs
@@ -538,7 +538,11 @@ pub mod pallet {
 			task_id
 		}
 
-		fn finish_task(network: NetworkId, task_id: TaskId, result: Result<(), ErrorMsg>) {
+		pub(crate) fn finish_task(
+			network: NetworkId,
+			task_id: TaskId,
+			result: Result<(), ErrorMsg>,
+		) {
 			TaskOutput::<T>::insert(task_id, result.clone());
 			if let Some(shard) = Some(TaskShard::<T>::take(task_id).unwrap()) {
 				log::debug!("finish task {task_id} on {shard}");
--- a/pallets/tasks/src/lib.rs
+++ b/pallets/tasks/src/lib.rs
@@ -471,7 +471,7 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
-		fn process_events(network: NetworkId, task_id: TaskId, events: GmpEvents) {
+		pub(crate) fn process_events(network: NetworkId, task_id: TaskId, events: GmpEvents) {
 			for event in events.0 {
 				match event {
 					GmpEvent::ShardRegistered(pubkey) => {
@@ -559,7 +559,7 @@ pub mod pallet {
 			Self::deposit_event(Event::TaskResult(task_id, result));
 		}
 
-		fn read_gateway_events(network: NetworkId) -> TaskId {
+		pub(crate) fn read_gateway_events(network: NetworkId) -> TaskId {
 			let block = SyncHeight::<T>::get(network);
 			let size = T::Networks::next_batch_size(network, block) as u64;
 			let end = block.saturating_add(size);
--- a/pallets/tasks/src/lib.rs
+++ b/pallets/tasks/src/lib.rs
@@ -543,6 +543,9 @@ pub mod pallet {
 			task_id: TaskId,
 			result: Result<(), ErrorMsg>,
 		) {
+			if TaskOutput::<T>::get(task_id).is_some() {
+				return;
+			}
 			TaskOutput::<T>::insert(task_id, result.clone());
 			if let Some(shard) = TaskShard::<T>::take(task_id) {
 				log::debug!("finish task {task_id} on {shard}");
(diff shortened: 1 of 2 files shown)


pallets/tasks/src/tests.rs
```diff
@@ -462,3 +462,51 @@ mod bench_helper {
...
+		// Assign the task to the shard
+		Tasks::assign_task(shard, task_id);
+		assert!(TaskShard::<Test>::contains_key(task_id), "Task should be assigned to a shard");
+		Tasks::finish_task(ETHEREUM, task_id, Ok(()));
+		assert!(!TaskShard::<Test>::contains_key(task_id));
+	});
+}
+
+#[test]
+fn test_regression_finish_task_panic() {
```
libreamartists:
```suggestion
fn test_regression_finish_task_does_not_panic() {
```

pallets/tasks/src/tests.rs
```diff
@@ -462,3 +462,51 @@ mod bench_helper {
...
+#[test]
+fn finish_task_removes_task_shard() {
+	new_test_ext().execute_with(|| {
+		register_gateway(ETHEREUM, 42);
+		let shard = create_shard(ETHEREUM, 3, 1);
+		roll(1);
+		let batch_id = BatchIdCounter::<Test>::get();
+		let task_id = Tasks::create_task(ETHEREUM, Task::SubmitGatewayMessage { batch_id });
+		BatchTaskId::<Test>::insert(batch_id, task_id);
+		// Assign the task to the shard
```
libreamartists:
```suggestion
```

peektism:
Not ready for review, not a full patch. This edge case also needs to solve the discrepancy of choosing whether the first or second result should be stored in TaskOutput. Current code emits two events for each submission but that does not make sense either. I think I would prefer to keep the first result rather than the second for this edge case because that is the result associated with processing events vs the one passed into submit_result is user submitted.

pallets/tasks/src/lib.rs
```diff
@@ -538,9 +538,16 @@ pub mod pallet {
 			task_id
 		}
 
-		fn finish_task(network: NetworkId, task_id: TaskId, result: Result<(), ErrorMsg>) {
+		pub(crate) fn finish_task(
+			network: NetworkId,
+			task_id: TaskId,
+			result: Result<(), ErrorMsg>,
+		) {
+			if TaskOutput::<T>::get(task_id).is_some() {
```
haider-rs: But if a task is re-executed and passed then this is a problem bcz it wont let the task store the Ok() result

libreamartists: It would already not store the Ok() result in the situation you described because there is already an early return in the submit_task_result extrinsic before we get to calling finish_task https://github.com/Analog-Labs/timechain/blob/22fe324d33f232d50e37bb048e2f3f6e36a13de1/pallets/tasks/src/lib.rs#L355

libreamartists: I added this additional return for the edge case where finish_task is called twice as a consequence of the same submit_task_result call in which case IMO we should keep the first result (from process_events) instead of overwriting it with the user submitted result

agryaznov (approved):

pallets/tasks/src/lib.rs
haider-rs: Thanks for pointing out

```suggestion
			if let Some(Ok(())) = TaskOutput::<T>::get(task_id) {
```
otherwise contains key does the trick right?

I also think the previous early return is also probelmatic for restarting tasks but i think the check is there for chronicle members to not again and again submit the same result which belongs to same shard 

i would also like input of @dvc94ch in here.

haider-rs: i mean for now PR looks good its just a normal flow for previous early return.

haider-rs (approved):

pallets/tasks/src/lib.rs
agryaznov: yep contains_key() would be better

libreamartists:
```suggestion
			if TaskOutput::<T>::contains_key(task_id) {
```

libreamartists: opened this issue as a follow up and changed it to contains_key

libreamartists: https://github.com/Analog-Labs/timechain/issues/1766

PR #1767, opened 2025-04-02
Replace get.is_some --> contains_key in pallets

peektism:
Replaces get().is_some() with contains_key() in all pallets, I think I got all the cases that exist.

This is an optimization that @agryaznov raised during another PR's review. It is an optimization because contains_key checks for existence without decoding the value while get() reads and decodes the value which is unnecessary when it is not being used in the logic thereafter


FlorianFranzen (approved):
Nice, I will make sure to update the benchmarks.

PR #1768, opened 2025-04-02
testnet: fix wrong keys in testnet genesis

penumbra23:
## Description

This PR fixes the testnet keys that I added by accident. #dontblamethedevops


FlorianFranzen (approved):
Looks good. 

Can not check the third council key, so that would be left to you.

PR #1807, opened 2025-04-15
Inner revert

### LINKED ISSUE #1803: run tests on real chains

dvc94ch:
Closes #1803 

Currently doesn't work due to faucet implementation

### DIFF
--- a/e2e-tests/src/lib.rs
+++ b/e2e-tests/src/lib.rs
@@ -15,7 +15,7 @@ use tempfile::TempDir;
 use testcontainers::{
 	core::{ContainerAsync, IntoContainerPort, Mount},
 	runners::AsyncRunner,
-	GenericImage, ImageExt,
+	ContainerRequest, GenericImage, ImageExt,
 };
 use tracing_subscriber::filter::EnvFilter;
 use zstd::{Decoder, Encoder};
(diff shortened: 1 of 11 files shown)


e2e-tests/src/lib.rs
```diff
@@ -199,6 +201,21 @@ impl TestEnvBuilder {
 		network: NetworkId,
 		shard_size: u16,
 		shard_threshold: u16,
+	) -> Result<()> {
+		let image = GenericImage::new("ghcr.io/foundry-rs/foundry", "latest")
+			.with_exposed_port(8545.tcp())
+			.with_env_var("ANVIL_IP_ADDR", "0.0.0.0").with_cmd([
+				"anvil -b=6 --steps-tracing --order=fifo --base-fee=0 --no-request-size-limit --slots-in-an-epoch 1 --state /state/anvil -s 7",
```
haider-rs, line 208: is block time 6 seconds necessary? wont it make test fast if we reduce it?

haider-rs (commented):
Thanks overall looks fine.

But the Astar image is just another local chain we are running test on. I think real chains should be testnet chains or mainnet chains deployed if wanna run on it. like forking using anvil and running it inside the docker container/or directly using their rpc endpoints.

https://book.getfoundry.sh/guides/forking-mainnet-with-cast-anvil

e2e-tests/src/lib.rs
dvc94ch: you can try, but I don't think it makes the test faster

dvc94ch:
we can already run tests on real chains, worked on that last week. we also need to test with the actual rpc/evm implementations to catch differences that we don't catch with just anvil. for example as we found out yesterday, astar is probably reverting the transaction because of the inner revert, something that doesn't happen with anvil.

PR #1810, opened 2025-04-21
Zenswap setup with tc-cli

haider-rs:
- [x] Adds swapping command in tc-cli.
- [x] Adds benchmarks for zenswap regular mode.
- [x] adds e2e test for zenswap


config/envs/development/config.yaml
```diff
@@ -104,6 +106,14 @@ networks:
...
     - "D2b0db0D3E48DeA780Fa6C417B015A0950D2A4A9"
     cctp_url: "https://iris-api-sandbox.circle.com/attestations/"
+    zenswap: 
+      universal_router: "3A9D48AB9751398BbFa63ad67599Bb04e4BdF98b"
+      permit2: "000000000022D473030F116dDEE9F6B43aC78BA3"
+      token_messenger: "9f3B8679c73C2Fef8b59B4f3444d4e156fb70AA5"
+      msg_transmitter: "7865fAfC2db2093669d92c0F33AeEF291086BEFD"
+      usdc: "1c7D4B196Cb0C7B01d743Fbc6116a902379C7238"
+      weth: "fFf9976782d46CC05630D1f6eBAb18b2324d6B14"
+
```
dvc94ch, line 114: where do these constants come from?

e2e-tests/src/lib.rs
```diff
@@ -113,6 +113,12 @@ impl TestEnvBuilder {
 								.join("analog-gmp/out/GatewayProxy.sol/GatewayProxy.json"),
 							gateway: workspace.join("analog-gmp/out/Gateway.sol/Gateway.json"),
 							tester: workspace.join("analog-gmp/out/GmpProxy.sol/GmpProxy.json"),
+							zenswap: Some(
+								workspace.join("analog-gmp/out/GmpProxy.sol/GmpProxy.json"),
+							),
+							zenswap_plugin: Some(
+								workspace.join("analog-gmp/out/GmpProxy.sol/GmpProxy.json"),
+							),
```
dvc94ch, line 126: doesn't seem right...

gmp/evm/src/dict.rs
```diff
@@ -34,3 +35,11 @@ pub(crate) fn load(data: &[u8]) -> Result<ChainDict> {
...
 	Ok(chains.into_iter().map(|c| (c.chain_id, c)).collect::<_>())
 }
+
+pub(crate) fn network_id_to_domain_id(network_id: NetworkId) -> Result<u32> {
+	match network_id {
+		10 => Ok(0),
+		13 => Ok(3),
+		_ => anyhow::bail!("Unsupported chain id"),
+	}
+}
```
dvc94ch: what's a domain id?

dvc94ch (commented):
looking good. still needs an e2e test that uses send_swap. looks like the only "important" parameter, the amount of the swap, isn't settable?

config/envs/development/config.yaml
haider-rs: Theese are the already deployed contracts by different organizations like circle, uniswap on chains. since we cant do swap without them there is no point of deploying ours.

e2e-tests/src/lib.rs
haider-rs: yeah just wanted to fix the compile time error. thanks will fix it

gmp/evm/src/dict.rs
haider-rs: thats the circle's cctp domain id that its contracts take. its like a network id for circle

e2e-tests/src/lib.rs
```diff
@@ -446,3 +474,32 @@ fn pick_free_port() -> Result<u16> {
...
+		}
+	}
+	pub fn to_config_backend(&self) -> Backend {
+		match self {
+			TestingBackend::Evm { .. } => Backend::Evm,
+			TestingBackend::Grpc => Backend::Grpc,
+		}
+	}
+	pub fn to_string(&self) -> String {
+		self.to_config_backend().to_string()
```
dvc94ch: that's not going to work for computing the snapshot name

dvc94ch: also you shouldn't implement to_string yourself, always implement std::fmt::Display

gmp/evm/src/sol.rs
```diff
@@ -148,6 +151,52 @@ sol! {
...
+		function swapSend(
+			bytes calldata pluginParams,
+			SwapParams calldata sourceParams,
+			SwapParams calldata destParams,
+			address payable recipient,
+			address plugin,
+			uint256 amountIn
+		) external payable;
+	}
+
```
dvc94ch: didn't we say we'd move this to the e2e test? its only needed there right?

tc-cli/src/lib.rs
```diff
@@ -1636,4 +1639,39 @@ impl Tc {
...
+		let blocks = self.read_events_blocks(exec, hash).await?;
+		let msgs = self.messages(dest, dest_addr, blocks).await?;
+		let msg = msgs
+			.into_iter()
+			.find(|msg| msg.message_id() == msg_id)
+			.context("failed to find message")?;
+		self.print_table(None, "message", vec![msg.clone()]).await?;
+		self.println(None, format!("received message after {} blocks", end - start))
+			.await?;
+		Ok(())
```
dvc94ch, line 1667: can you use this track_msg_id in message_trace? seems like its duplicate code. this is message_trace without listing the messages at the end right?

tc-cli/src/swap_benchmark.rs
```diff
@@ -0,0 +1,163 @@
...
+							error_count = 0;
+					},
+						Err(e) => {
+							tracing::error!("Error sending swap, {:?}", e);
+							error_count += 1;
+						},
+					};
+				}
+			}
+		}
```
dvc94ch, line 160: what's the difference to the other benchmark? why is it duplicated?

gmp/evm/src/sol.rs
haider-rs: sorry moved.

tc-cli/src/lib.rs
haider-rs: no this is like the wait we do in the smoke-test which calls the message_trace for block inside it

tc-cli/src/swap_benchmark.rs
haider-rs: the difference this benches only runs for swaps and between two chains. Other benchmarks runs for all the chains and send message across each other. while this just benchmarks from chain a to chain b with swap calls.

haider-rs:
No longer supports CCTP.

PR #1819, opened 2025-04-27
delegated-staking: do not reap pool stakers on full withdrawal

FlorianFranzen:
## Description

Fixes upstream bug that caused user on mainnet to be reaped when trying to withdraw all their staked tokens.


FlorianFranzen:
Will be included in #1863 to ensure save deployment.

Issue #1818, opened 2025-04-25
PolkaVM openzeppelin compatibility test suite


peektism:
Follow up for #1740 to add E2E tests

PR #1740, opened 2025-03-25
PolkaVM MVP

peektism:
Closes #1729

## TODO
- [x] upgrade to https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2503 (released April 9th+)
- [x] fix metadata generation by moving `runtime/src/api.rs --> runtime/src/lib.rs`
- [x] **pallet_revive** configured for testnet runtime to follow conventions set in the Westend AssetHub configuration (`polkadot-sdk/cumulus/parachains/assets/westend-asset-hub/src/lib.rs`) see https://github.com/Analog-Labs/timechain/pull/1740#issuecomment-2822158838
- [x] impl ETH RPC + RuntimeApis https://github.com/Analog-Labs/timechain/pull/1804
- [x] e2e tests pass
- [x] migrations pass try-runtime check cc @FlorianFranzen 
- [x] [demo using local node](https://www.notion.so/teamanalog/PolkaVM-MVP-Demo-Steps-1e04872af8ca8056ac5bc586760e1af4?pvs=4) 

## Follow Ups
- https://github.com/Analog-Labs/timechain/issues/1748
- https://github.com/Analog-Labs/timechain/issues/1818
- https://github.com/Analog-Labs/timechain/issues/1817

PR #1822, opened 2025-04-28
PolkaVM E2E Tests

peektism:
WIP Addresses #1818 

Based on top of #1740 until it is merged


peektism:
Having trouble figuring out how to add a second process for running the revive-ethereum-rpc. Thinking about it still.

peektism:
not the correct approach, working on the right way to add this

PR #1834, opened 2025-05-05
Remove target block stream.

dvc94ch:
No issues, was straightforward to remove.


haider-rs (approved):
looks good, let see if e2e tests catch any error!

PR #1839, opened 2025-05-06
Register route fix

dvc94ch:


haider-rs:
can we please also use same coin id as network 10 for network 13?

PR #1843, opened 2025-05-07
Fix circle tls issue

haider-rs:


dvc94ch:
doubt we're hitting the rate limit or are we?

haider-rs:
no probably not. bcz otherwise it could have return 429 status code and we are just sending a single request. Plus doesnt really need to wait for 35s for each task. And add proper limiters to improve effeciency.

dvc94ch (approved):

PR #1846, opened 2025-05-08
Minimize gas costs for PolkaVM

peektism:
These changes should resolve the issues with contract deployment testing that @aliTanveerAnalog is experiencing. We expect that the issues are due to bug(s) in gas estimation in the currently used version of pallet-revive.

## Follow Up(s)

 If we decide to merge this PR, we need to make a follow up issue so that these values are adjusted before mainnet launch so that we do not accidentally expose a DDOS attack vector.

Currently investigating a long-term fix that will be more secure than this short term patch. This research is driven by reading [differences with PolkaVM Gas Configuration](https://contracts.polkadot.io/differences_to_eth#scaling-of-gas-values) and the pallet-revive source code.

### DIFF
--- a/runtime/src/configs/revive.rs
+++ b/runtime/src/configs/revive.rs
@@ -9,11 +9,21 @@ use sp_runtime::{
 	Perbill,
 };
 
-// Temporarily set to low values as a workaround for known gas estimation issues
+// Temporarily set to low deposit requirements for integration environment
+// as a workaround for known gas estimation issues
 parameter_types! {
-	pub const DepositPerItem: Balance = 1;//deposit(1,0);
-	pub const DepositPerByte: Balance = 1;//deposit(0,1);
-	pub CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(0);//Perbill::from_percent(30);
+	#[cfg(feature = "develop")]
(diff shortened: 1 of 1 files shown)


runtime/src/configs/revive.rs
```diff
@@ -9,10 +9,11 @@ use sp_runtime::{
 	Perbill,
 };
 
+// Temporarily set to low values as a workaround for known gas estimation issues
 parameter_types! {
-	pub const DepositPerItem: Balance = deposit(1, 0);
-	pub const DepositPerByte: Balance = deposit(0, 1);
```
libreamartists, line 14: Do our deposit functions return significantly higher values than the Westend AssetHub runtime deposit function?

libreamartists: That may be the reason for higher than expected storage deposit requirements for contract deployment.

libreamartists: Did not notice a significant difference so I don't think that is the problem

peektism:
The problem was just an eth rpc bug that exists for eth_getTransactionReceipt so this solution was not necessary. Gas costs still seemed higher than I would expect. They are for a parachain so it is possible we could make them cheaper and be fine.

PR #1867, opened 2025-05-13
launch: add investor snapshot 1

FlorianFranzen:
This add four more launch stages to the chain in accordance with the first investor snapshot.


taniasaleem (approved):
Checked the wallets and amount, everything looks good

PR #1875, opened 2025-05-15
Generalize evm backend

dvc94ch:


gmp/evm/src/lib.rs
```diff
@@ -302,179 +278,100 @@ impl IConnector for Connector {
...
+			.deploy_contract(
+				gateway,
+				Gateway::constructorCall {
+					network: self.network_id,
+					proxy,
+				},
+			)
+			.await?;
+
+		let call = GatewayProxy::upgradeCall {
```
haider-rs, line 322: giving error of upgradeCall not found here.

primitives/src/gmp.rs
```diff
@@ -365,15 +365,10 @@ pub trait IConnector: IChain {
 #[cfg(feature = "std")]
 #[async_trait::async_trait]
 pub trait IConnectorAdmin: IConnector {
+	/// Deploys the proxy contract.
+	async fn deploy_proxy(&self, proxy: &[u8]) -> Result<(Address32, u64)>;
```
haider-rs, line 369: not all chains needs this. shouldnt this be done in the deploy_gateway since its evm sepcific? Solana smart contracts are upgradable by default.

haider-rs (commented):
looks good! also dropped a comment on analog-gmp

primitives/src/gmp.rs
dvc94ch: I guess we can just leave the implementation empty for solana

gmp/evm/src/lib.rs
dvc94ch: yep, that's because its using the old analog-gmp

primitives/src/gmp.rs
haider-rs: but then this will be probelmatic for tc-cli:
https://github.com/Analog-Labs/timechain/pull/1875/files#diff-fa872fcf6771b29bb6bfc7febf71ac2c3301dfce0764317a13d6b5318032042eR862

haider-rs:
also i think the deploy and redeploy should be seperate functions. bcz other chains like solana have seperate logic for both.

dvc94ch:
merging for now, will follow up with your concerns @haider-rs

PR #1878, opened 2025-05-19
Improves tc-cli gmp benchmarks

haider-rs:


tc-cli/src/lib.rs
```diff
@@ -1118,7 +1118,7 @@ impl Tc {
 			gateways.insert(network, gateway);
 		}
 
-		self.register_routes(gateways).await?;
+		// self.register_routes(gateways).await?;
```
dvc94ch: why was this needed?

dvc94ch (commented):
If you rebase I'll review it tomorrow

tc-cli/src/lib.rs
haider-rs: the reason was that this call requires admin access to gateway while i was trying on integration and not wanted to use admin keys thats why i commented it however now i have ran on locally so i removed this commented code.

dvc94ch (approved):

dvc94ch (approved):

PR #1900, opened 2025-05-28
polkadot-sdk: v1.18.3-anlog0 -> v1.18.5-anlog0

FlorianFranzen:
Updates to latest Polkdot SDK,, most notably has some fixes for revive tracing and scheduler pallet.


FlorianFranzen:
@penumbra23 Do we need to do anything to support the new RPC node after we merge this?

penumbra23:
@FlorianFranzen Need to inspect the diff from last deployment, but I thunk @4meta5 is almost done with Postgres integration, so either way we'll need to redeploy. For now it's fine since we use the prebuilt images.

PR #1918, opened 2025-06-09
polkadot-sdk: v1.18.5 -> v1.18.6

FlorianFranzen:
This updates us to the latest polkadot-sdk, improving some of the networking.


dvc94ch:
Can we stop building libp2p too? It's a heavy dependency

FlorianFranzen:
> Can we stop building libp2p too? It's a heavy dependency

Unless there are some changes to `sc-network`, it will not be that easy to remove the `libp2p` dependency.

PR #1926, opened 2025-06-11
Implement lazy connector.

dvc94ch:


haider-rs (approved):

agryaznov (approved):

FlorianFranzen (approved):
Looks reasonable, but only gave it a quick review. Clippy also still seems to be unhappy.

PR #1929, opened 2025-06-12
Adds max gas price in network pallet

haider-rs:


e2e-tests/src/lib.rs
```diff
@@ -257,6 +258,7 @@ impl TestEnvBuilder {
 				msg_session_gas: 50_000,
 				currency_decimals: 18,
 				currency_symbol: "ETH".into(),
+				max_gas_price: 0,
```
dvc94ch: this won't work, the max_gas_price needs to be at least 1 no?

dvc94ch (approved):

PR #1931, opened 2025-06-13
Check gas price before submitting tx from chronicle

haider-rs:


gmp/evm/src/lib.rs
```diff
@@ -244,6 +244,11 @@ impl IConnector for Connector {
 		self.submit(tx).await.map_err(|err| err.to_string())?;
 		Ok(())
 	}
+
+	/// Get EIP1559 `max_fee_per_gas` estimate for the connector's chain
+	async fn max_fee_per_gas(&self) -> Result<u128> {
+		self.estimate_eip1559_fees().await.map(|e| e.max_fee_per_gas)
+	}
```
dvc94ch:
```suggestion
	async fn gas_price(&self) -> Result<u128> {
		self.estimate_eip1559_fees().await.map(|e| e.max_fee_per_gas + e.max_priority_fee)
	}
```

tc-cli/src/main.rs
```diff
@@ -300,14 +297,6 @@ async fn real_main() -> Result<()> {
...
 			tc.print_table(None, "pending-batches", batches).await?;
 		},
-		Command::MaxFeePerGas { network } => {
-			let fee = tc.max_fee_per_gas(network).await?;
-			tc.println(
-				None,
-				format!("EIP1559 max_fee_per_gas for network: {} is : {}", network, fee),
-			)
-			.await?;
-		},
```
dvc94ch, line 310: could have left it, but is ok, will build a better command that shows more data

chronicle/src/tasks/mod.rs
```diff
@@ -146,16 +146,27 @@ impl TaskParams {
...
-				{
-					tracing::error!(parent: &span, "Error while executing batch: {e}");
-					e.truncate(time_primitives::MAX_ERROR_LEN as usize - 4);
-					let result = TaskResult::SubmitGatewayMessage {
-						error: ErrorMsg(BoundedVec::truncate_from(e.encode())),
-					};
-					tracing::debug!(parent: &span, "submitting task result");
-					self.runtime.submit_task_result(task_id, result).await?;
+				let max_gas_price =
+					self.runtime.get_network_gas_price(network_id, block_hash).await?;
```
dvc94ch:
```suggestion
					self.runtime.network_gas_price(network_id, block_hash).await?;
```

chronicle/src/tasks/mod.rs
```diff
@@ -146,16 +146,27 @@ impl TaskParams {
...
+				let current_gas_price = self.connector.max_fee_per_gas().await?;
```
dvc94ch: this can probably be moved up to the top loop, no need to query it for every task, just need to query it once per block

chronicle/src/tasks/mod.rs
```diff
@@ -146,16 +146,27 @@ impl TaskParams {
...
+				if current_gas_price <= max_gas_price {
+					tracing::info!(parent: &span, "current_gas price: {current_gas_price} <= max_gas_price: {max_gas_price}");
+					if let Err(mut e) = self
+						.connector
+						.submit_commands(gateway, batch_id, msg, signer, signature)
```
dvc94ch: should include the gas price computed at top of loop.

chronicle/src/tasks/mod.rs
```diff
@@ -146,16 +146,27 @@ impl TaskParams {
...
-					let result = TaskResult::SubmitGatewayMessage {
-						error: ErrorMsg(BoundedVec::truncate_from(e.encode())),
-					};
-					tracing::debug!(parent: &span, "submitting task result");
-					self.runtime.submit_task_result(task_id, result).await?;
+				let max_gas_price =
+					self.runtime.get_network_gas_price(network_id, block_hash).await?;
+				let current_gas_price = self.connector.max_fee_per_gas().await?;
+				if current_gas_price <= max_gas_price {
+					tracing::info!(parent: &span, "current_gas price: {current_gas_price} <= max_gas_price: {max_gas_price}");
```
dvc94ch: notify the admin interface somehow

gmp/evm/src/lib.rs
dvc94ch: best to return a struct here and add them in the connector

haider-rs: it already contians the max_fee + priority_fee:
```
pub fn eip1559_default_estimator(
    base_fee_per_gas: u128,
    rewards: &[Vec<u128>],
) -> Eip1559Estimation {
    let max_priority_fee_per_gas = estimate_priority_fee(rewards);
    let potential_max_fee = base_fee_per_gas * EIP1559_BASE_FEE_MULTIPLIER;

    Eip1559Estimation {
        max_fee_per_gas: potential_max_fee + max_priority_fee_per_gas,
        max_priority_fee_per_gas,
    }
}

```

chronicle/src/tasks/mod.rs
```diff
@@ -108,56 +116,65 @@ impl TaskParams {
...
+						let result = TaskResult::SubmitGatewayMessage {
+							error: ErrorMsg(BoundedVec::truncate_from(e.encode())),
+						};
+						tracing::debug!(parent: &span, "submitting task result");
+						self.runtime.submit_task_result(task_id, result).await?;
+					}
+				},
+			}
+		} else {
+			tracing::warn!(parent: &span, "current_gas price: {current_gas_price} > max_gas_price: {max_gas_price}");
```
dvc94ch: this doesn't work, it will never complete

chronicle/src/tasks/mod.rs
```diff
@@ -108,56 +116,65 @@ impl TaskParams {
...
-						error: ErrorMsg(BoundedVec::truncate_from(e.encode())),
-					};
-					tracing::debug!(parent: &span, "submitting task result");
-					self.runtime.submit_task_result(task_id, result).await?;
-				}
-			},
+		let max_gas_price = self.runtime.network_gas_price(network_id, block_hash).await?;
+		let current_gas_price = self.gas_price().await?;
+		if current_gas_price <= max_gas_price {
+			tracing::info!(parent: &span, "current_gas price: {current_gas_price} <= max_gas_price: {max_gas_price}");
```
dvc94ch: this needs to be moved to executor. this fetches the gas price for every task which is pointless. the task should not get started

chronicle/src/mock.rs
```diff
@@ -198,6 +198,10 @@ impl Runtime for Mock {
 			.map(|network| network.chain_name.clone()))
 	}
 
+	async fn network_gas_price(&self, _: NetworkId, _: BlockHash) -> Result<u128> {
+		Ok(100)
+	}
+
 	async fn get_member_peer_id(
```
dvc94ch: thought I removed all the `get_` prefixes once. they are not really needed, in rust we usually don't prefix getters with `get_`

chronicle/src/tasks/mod.rs
```diff
@@ -187,24 +198,28 @@ impl TaskExecutor {
...
-		let tasks = self.params.runtime.get_shard_tasks(shard_id, block_hash).await?;
+		let tasks = self.params.runtime.shard_tasks(shard_id, block_hash).await?;
 
 		let failed_tasks: Arc<Mutex<u64>> = Default::default();
+
+		let max_gas_price =
+			self.params.runtime.network_gas_price(self.params.network(), block_hash).await?;
+		let current_gas_price = self.params.gas_price().await?;
 		for task_id in tasks.iter().copied() {
+			if current_gas_price > max_gas_price {
```
haider-rs, line 213: the reason for including in the loop is to complete the flow of the function otherwise in high gas spikes the execution of completing task would stop

dvc94ch (approved):

